PR #8, opened 2020-08-14
remove repeaed keywords

paizzj:
'.' keyworkd is redundancy


arvidn:
I believe this patch changes the op-code values, so this is not sufficient to keep things working. I think the first `.` should be removed from that list too. afaik, they are there to preserve the existing opcode values.

The plan is to compact these before the final release.

richardkiss:
That's correct, the `.` is just a placeholder.

PR #9, opened 2020-10-12
Add print

aqk:
This commit intentionally does not add an op_print opcode.
    The print atom is intended for debugging in a development environment.
    print passes through the value of its reduced arguments, so it can be inserted anywhere in the tree.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install .[dev]
+	  python -m pip install .[dev]
+          python -m pip install pytest
       - name: Test with pytest
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -22,7 +22,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install .[dev]
+	  python -m pip install .[dev]
+          python -m pip install pytest
```
richardkiss, line 26: I believe `pip install .[dev]` already installs `pytest`, so this isn't really necessary.

aqk: I don't think it did in my forked repo, but I could be mistaken

aqk:
Closing this PR in preference to an all-clvm_tools implementation.

PR #10, opened 2020-11-18
there is no division operator

arvidn:
(only divmod), so remove the rewrite entry


richardkiss:
Cherry-picked as d174cd927f28e7c0c927587010bad444f6b6fd5c

PR #11, opened 2020-11-18
move cost constants into costs.py

arvidn:


richardkiss:
Cherry-picked as 16889b1ed5f10af3f72dc55d9405c01bce025ace.

PR #14, opened 2020-12-04
remove sha256tree operator

arvidn:

PR #16, opened 2020-12-07
New costs

arvidn:
this depends on https://github.com/Chia-Network/clvm/pull/14


clvm/more_ops.py
```diff
@@ -156,7 +201,7 @@ def op_strlen(args):
     if a0.pair:
         raise EvalError("len on list", a0)
     size = len(a0.as_atom())
-    cost = size
+    cost = STRLEN_BASE_COST + size / STRLEN_COST_PER_BYTE_DIVIDER
```
richardkiss, line 204: Do you mean `//` here?

arvidn: ah, yes, I did

richardkiss:
The divisions concern me a little bit, as I worry they will overwhelm the operations they are calculating the cost of. I do notice that they all tend to be powers of 2... what do you think about changing these to `>>`, which we know will run quickly.

arvidn:
is there any doubt that `// 64` is not identical to `>> 6`? I think the former is much more legible

richardkiss:
There's no doubt, I'm just concerned it's slower. Probably C or rust would optimize it, but I'm not sure that python does (since it has no assurances that what we humans consider a constant doesn't actually change).

arvidn:
in C and C++ I would say `/` is preferred over `>>` because there are certain edge-cases where `>>` behaves unexpected, especially on signed variables.

arvidn:
I rebased and fixed the conflicts with the `args-**` changes. I also fixed the floating-point division -> integer division. I applied that as a separate commit to ease review, but I think they should be squashed before landing.

richardkiss:
```(venv) kiss@oasis develop % python -m timeit -s "T = 10004843716274856703848" "T >> 6"
5000000 loops, best of 5: 59.1 nsec per loop
(venv) kiss@oasis develop % python -m timeit -s "T = 10004843716274856703848" "T // 64"
5000000 loops, best of 5: 87.1 nsec per loop
(venv) kiss@oasis develop % python -m timeit -s "T = 100" "T // 64" 
10000000 loops, best of 5: 34.6 nsec per loop
(venv) kiss@oasis develop % python -m timeit -s "T = 100" "T >> 6" 
5000000 loops, best of 5: 59.1 nsec per loop
```

Our values on the left are likely to be closer to 100 than 10004843716274856703848, so turns out that `//` is within order of magnitude even in the worst case. My intuitions were off here.

PR #17, opened 2020-12-08
add tests for edge cases for all operators

arvidn:
please review the test cases and the expected output. I did and discovered a few missing checks that I'm adding in this patch.


tests/edge-cases/point_add-05.txt
```diff
@@ -0,0 +1,5 @@
+brun '(point_add (q (1 2)) (q (1 2)))'
+FAIL: point_add expects blob, got ff01ff0280: from_bytes(): incompatible function arguments. The following argument types are supported:
+    1. (self: buffer) -> blspy.G1Element
+
+Invoked with: None ((q 2) (q 2))
```
arvidn, line 5: I'm not sure about this one. Is this correct? the output doesn't follow the pattern used by other operator failures, which I've learned sometimes indicate the wrong kind of exception is thrown (i.e. an unexpected exception)

tests/edge-cases/apply-02.txt
```diff
@@ -0,0 +1,2 @@
+brun '(a (q 1))'
+FAIL: unimplemented operator 3
```
arvidn, line 2: this is a bit questionable as well, but it seems odd to have `a` in the operator list at all, before it's implemented

clvm/more_ops.py
```diff
@@ -152,15 +152,19 @@ def op_point_add(items):
 
 
 def op_strlen(args):
+    if args.list_len() != 1:
+        raise EvalError("strlen takes exactly 1 argument", args)
     a0 = args.first()
     if a0.pair:
-        raise EvalError("len on list", a0)
+        raise EvalError("strlen on list", a0)
```
arvidn, line 159: this affects one of the existing test cases. I think `strlen` is better than `len` here.

tests/edge-cases/if-07.txt
```diff
@@ -0,0 +1,2 @@
+brun '(i (q (1 2)) (q (1 2)) (q (1 2)))'
+(q 2)
```
arvidn, line 2: Is i intentional to treat a (non-empty) list as `true`?

tests/edge-cases/logand-01.txt
```diff
@@ -0,0 +1,2 @@
+brun '(logand)'
+-1
```
arvidn, line 2: @aqk this behavior is documented, right?

tests/edge-cases/lognot-04.txt
```diff
@@ -0,0 +1,2 @@
+brun '(lognot ())'
+-1
```
arvidn, line 2: @aqk and this behavior, it is not necessarily obvious

tests/edge-cases/point_add-06.txt
```diff
@@ -0,0 +1,2 @@
+brun '(point_add (q 0xffff) (q 0xffff))'
+FAIL: point_add expects blob, got 82ffff: Length of bytes object not equal to G1Element::SIZE (0xffff 0xffff)
```
arvidn, line 2: it's a little bit odd that the error message prints the *serialized* form of the value it was passed

clvm/core_ops.py
```diff
@@ -47,6 +47,8 @@ def op_raise(args):
 
 
 def op_eq(args):
+    if args.list_len() != 2:
```
matt-o-how, line 50: I thought that we wanted equal to take unlimited arguments so you could assert equality between multiple things without needing multiple checks?

arvidn: perhaps. that's not what it does now though. If we make that change, we can change this check too.

arvidn: I'm not sold on that being desirable though, at least not in the low-level language. There's consistency with `>` and `>s` right now (in being binary operators).

arvidn:
@aqk @richardkiss ping

tests/edge-cases/if-07.txt
richardkiss: Yes

tests/edge-cases/point_add-05.txt
richardkiss: Yeah, probably worth checking the pair/atom type and raising a different message prior to trying to evaluate.

tests/edge-cases/point_add-06.txt
richardkiss: Hmm, that is odd. Probably could do with a realignment.

tests/edge-cases/softfork-02.txt
```diff
@@ -0,0 +1,2 @@
+brun '(softfork (q 1))'
+()
```
richardkiss, line 2: Maybe the `softfork` tests should include `-c` so costs are output.

tests/edge-cases/point_add-06.txt
arvidn: when an `SExp` is printed (via `__str__`) it uses `to_bin()` which in turn uses the serializer. I think it would be better to keep the serialization separate. Ideally we could change the serialization format without changing the way an `SExp` is printed. I will propose a patch for this

arvidn: I've made this change, and removed `to_bin()`. Please take a look

tests/edge-cases/point_add-05.txt
arvidn: done, please review

arvidn:
I've updated this PR with additional commits to make it easier to review. I intend to squash once it's in good shape

clvm/SExp.py
```diff
@@ -170,7 +164,7 @@ def as_python(self):
         return as_python(self)
 
     def __str__(self):
-        return self.as_bin().hex()
+        return self.atom.hex()
```
arvidn, line 167: this won't work for cons objects

richardkiss: Yes, seems like this will change test outputs, as `as_bin` did serialization.

clvm/core_ops.py
richardkiss: Maybe I'm paranoid, but I am also afraid that `=` with multiple values could lead to subtle consensus bugs across implementations.

clvm/SExp.py
arvidn: the intention was to change the test output, or rather, the error messages. But when we actually serialize things with the serializer, we don't use `as_bin()`, we use the serializer. All tests still pass with this change, it's just there is no test for this `__str__` function yet, and it would fail for non-atoms.

tests/edge-cases/divmod-01.txt
```diff
@@ -0,0 +1,2 @@
+brun '(divmod)'
```
aqk, line 1: @arvidn @richardkiss If we only return one value, we should consider changing the name `divmod` -> e.g. `div`

tests/edge-cases/divmod-09.txt
```diff
@@ -0,0 +1,2 @@
+brun '(divmod (q -1) (q -1))'
```
aqk, line 1: We should check all sign combinations. One of those weird int math corners.

https://www.evanjones.ca/integer-division.html

tests/edge-cases/list-04.txt
```diff
@@ -0,0 +1,2 @@
+brun '(l ())'
```
aqk, line 1: This is a good one. I mention this in the doc, but I'll try to mention it several times.

tests/edge-cases/logand-01.txt
aqk: Yes, it's in the table of outputs with zero arguments, and in the arithmetic unity table, but there it's described as "all ones", so I'll improve that.

tests/edge-cases/logand-05.txt
```diff
@@ -0,0 +1,2 @@
+brun '(logand (q (1 2)) (q (1 2)))'
+FAIL: logand requires int args ((q 2) (q 2))
```
aqk, line 2: This error message would be hard to understand if I hadn't spent so much time with clvm. I know this is not new in your patch.

tests/edge-cases/logand-07.txt
```diff
@@ -0,0 +1,2 @@
+brun '(logand (q 128) (q 128))'
```
aqk, line 1: For the logical operators, I feel like it would be good to include a value that spans one extra byte, like 0x01ff

tests/edge-cases/lognot-02.txt
```diff
@@ -0,0 +1,2 @@
+brun '(lognot (q 1))'
+-2
```
aqk, line 2: Making a note to explain this explicitly & well in the docs.

tests/edge-cases/lognot-06.txt
```diff
@@ -0,0 +1,2 @@
+brun '(lognot (q 0xffff))'
```
aqk, line 1: Nice one.

tests/edge-cases/lsh-01.txt
```diff
@@ -0,0 +1,2 @@
+brun '(lsh)'
+FAIL: ash requires 2 args ()
```
aqk, line 2: Hold the phone. opcode is `lsh`, but message says `ash`

tests/edge-cases/not-05.txt
```diff
@@ -0,0 +1,2 @@
+brun '(not (q (1 2)))'
```
aqk, line 1: `brun '(not (q (1 2)))'`
@richardkiss This seems like it should be an error to me.

clvm/more_ops.py
aqk: I agree. `strlen` is more direct.

tests/edge-cases/all-01.txt
```diff
@@ -0,0 +1,2 @@
+brun '(all)'
+1
```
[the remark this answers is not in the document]
matt-o-how: Why not enforce that it has at least 1 parameter? I can't think of a circumstance where you would want to call it without any parameters.

clvm/core_ops.py
aqk: I agree, better to be conservative and only allow two arguments at the lowest level.

tests/edge-cases/all-01.txt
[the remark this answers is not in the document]
aqk: Yes, that's what I'm saying too. It's not useful with zero args.

tests/edge-cases/apply-02.txt
aqk: No problem. This documents that it doesn't work in this branch.
I think that's good.

tests/edge-cases/all-01.txt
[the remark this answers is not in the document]
richardkiss: Here's one I can think of: when used with the new `apply`, you might do `(a #all ITEMS)`; it would save you checking against ITEMS being zero. Note that in python, `any` and `all` each accept zero parameters.

tests/edge-cases/ash-02.txt
```diff
@@ -0,0 +1,2 @@
+brun '(ash (q 1))'
+FAIL: ash requires 2 args (q)
```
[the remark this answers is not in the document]
richardkiss: One option is the s-expression in the error message can be decompiled without keywords (since it generally just includes already-evaluated expressions).

tests/edge-cases/divmod-01.txt
richardkiss: This hasn't been changed yet, as we haven't come to a consensus on `div` vs `divmod`. Let me run a quick test.

richardkiss:
```
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "divmod(A,B)"
10000000 loops, best of 3: 0.164 usec per loop
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "A/B"        
10000000 loops, best of 3: 0.0503 usec per loop
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "A%B"
10000000 loops, best of 3: 0.0497 usec per loop
```

Okay, this is pretty wack. I don't know what to make of these results.

richardkiss: It's almost as if it's spending most of its time making the tuple.

richardkiss: Okay, let's use integer division.
```
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "A//B"       
10000000 loops, best of 3: 0.0522 usec per loop
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "A%B" 
10000000 loops, best of 3: 0.0512 usec per loop
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "(A//B, A%B)"
10000000 loops, best of 3: 0.117 usec per loop
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "divmod(A,B)"
10000000 loops, best of 3: 0.166 usec per loop
kiss@oasis ~ % python -m timeit 'A = 74738383914845; B = 83717484' "Q = A//B; (Q, A - B * Q)"
10000000 loops, best of 3: 0.104 usec per loop
```

Seems like calculating the remainder using arithmetic is the fastest. I recommend we add integer division and keep divmod. Your thoughts @aqk @arvidn @bramcohen ?

tests/edge-cases/logand-05.txt
richardkiss: Maybe we should print just the first invalid argument instead of all of them.

tests/edge-cases/not-05.txt
richardkiss: This is a boolean operator, so it ought to have the same semantics as `i`. This is how `i` works too.

tests/edge-cases/all-01.txt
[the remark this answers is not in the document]
aqk: Yes, makes sense if you are generating params with other code.

tests/edge-cases/divmod-01.txt
aqk: I'm ok with keeping both, as long as div(A,B) always means exactly divmod(A,B)[0]

clvm/SExp.py
arvidn: I'm backing all of this out. I will propose a less intrusive alternative as a separate PR

tests/edge-cases/concat-04.txt
```diff
@@ -0,0 +1,2 @@
+brun '(concat () ())'
+()
```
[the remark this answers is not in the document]
arvidn: added as `concat-09.txt` and `concat-10.txt`

tests/edge-cases/concat-05.txt
```diff
@@ -0,0 +1,2 @@
+brun '(concat (q (1 2)) (q (1 2)))'
+FAIL: concat on list (q 2)
```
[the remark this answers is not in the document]
arvidn: added as `concat-11.txt` and `concat-12.txt`

arvidn:
I've rebased this on top of develop. please make sure the changes in test cases are correct!

https://github.com/Chia-Network/clvm/pull/17/commits/0f5673466ca5960ba64d739ffa5481d7f2028309

I dropped the changes to `as_bin()` and `__str__`.

I believe the remaining comments are about `divmod` and `div`, which should be a separate PR. It's risky to rebase this since the existing (and reviewed) test cases may get mixed up in new changes that didn't have tests when they landed, so I think we should be careful to not widen the scope of this PR.

arvidn:
I still think it makes sense to squash this after review, but I don't have strong feelings about it if it may delay landing it

PR #18, opened 2020-12-09
Quote as cons pair

aqk:


aqk:
Note that this needs https://github.com/Chia-Network/clvm/tree/quote-as-cons-pair

aqk:
https://github.com/Chia-Network/clvm/pull/28

PR #19, opened 2020-12-14
enable test coverage in CI

arvidn:


.github/workflows/main.yml
```diff
@@ -23,6 +23,17 @@ jobs:
...
           python -m pip install --upgrade pip
           python -m pip install .[dev]
+          python -m pip install coverage
+          python -m pip install coveralls
       - name: Test with pytest
         run: |
           py.test tests
+          coverage run --branch --include=clvm/SExp.py -m pytest tests/as_python_test.py
+          coverage run -a --branch --include=clvm/serialize.py -m pytest tests/serialize_test.py
+          coverage run -a --branch --source=clvm --omit=clvm/SExp.py,clvm/serialize.py -m pytest tests
```
arvidn, line 33: instead of just running all tests and merging all test coverage, I specifically *just* measure test coverage of `SExp.py` based on the deliberate tests written for it, in `as_python_test.py`. Same thing with `serialize.py`. This is to avoid "collateral" coverage by unrelated tests that don't necessarily validate the result of that particular code being executed. I think this gives a more honest representation of what's being tested.

tests/as_python_test.py
```diff
@@ -30,3 +57,73 @@ def test_complex(self):
...
+        # list
+        self.assertEqual(SExp.to(b'\x01').cons(SExp.to(b'\x02').cons(SExp.null())).as_python(), [b'\x01', b'\x02'])
+        # cons-box of two values
+        self.assertEqual(SExp.to(b'\x01').cons(SExp.to(b'\x02').as_python()), (b'\x01', b'\x02'))
+
+    def test_deep_recursion(self):
+        d = [b"2"]
+        for i in range(480):
+            d = [d]
+#        self.check_as_python(d)
```
arvidn, line 103: this fails, and it's not trivial to fix. there's a recursive function that probably should be made non-recursive

arvidn:
I started adding more unit tests to increase the coverage, but I think I should do that in separate PRs

arvidn:
this is the current coverage report: https://coveralls.io/builds/35794098

richardkiss (approved):

PR #20, opened 2020-12-15
remove unused validate() function on SExp class

arvidn:
It's also recursive which makes it a bit risky to leave around, in case it's used on untrusted input


richardkiss:
Cherry-picked as 9ecf1f71cd8dd6a1a90bd24eb5614e18ded8f171.

PR #21, opened 2020-12-15
redundant SExp.to() call

arvidn:
remove this call as it appears to be redundant

### DIFF
--- a/clvm/SExp.py
+++ b/clvm/SExp.py
@@ -112,7 +112,7 @@ class SExp(CLVMObject):
         return class_(v1)
 
     def cons(self, right: "CLVMObject"):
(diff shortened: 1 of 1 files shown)


clvm/SExp.py
```diff
@@ -112,7 +112,7 @@ def to(class_, v: CastableType):
         return class_(v1)
 
     def cons(self, right: "CLVMObject"):
```
arvidn, line 114: as I understand it, `right` will always be  CLVM object here, and can be passed directly into the right hand side of the cons-box

richardkiss:
Fixed in 0ba63c13798951c241e529d67d2f4e6b8ac862b7 (I had this patch laying around in another branch).

PR #22, opened 2020-12-17
100% coverage of serialize.py

arvidn:
extend unit test of serialize.py to cover all of it

### DIFF
--- a/clvm/serialize.py
+++ b/clvm/serialize.py
@@ -124,6 +124,8 @@ def _atom_from_stream(f, b, to_sexp):
             raise ValueError("bad encoding")
         size_blob += b
     size = int.from_bytes(size_blob, "big")
+    if size >= 0x400000000:
+        raise ValueError("blob too large")
     blob = f.read(size)
     if len(blob) != size:
         raise ValueError("bad encoding")
--- a/tests/serialize_test.py
+++ b/tests/serialize_test.py
@@ -7,6 +7,17 @@ from clvm.serialize import (sexp_from_stream, atom_to_byte_iterator)
 
 TEXT = b"the quick brown fox jumps over the lazy dogs"
 
+class InfiniteStream(io.TextIOBase):
+    def __init__(self, b):
+        self.buf = b
+    def read(self, n):
+        ret = b''
+        while n > 0 and len(self.buf) > 0:
+            ret += self.buf[0:1]
+            self.buf = self.buf[1:]
+            n -= 1
+        ret += b' ' * n
+        return ret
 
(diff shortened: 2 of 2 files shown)


tests/serialize_test.py
```diff
@@ -44,6 +44,12 @@ def test_long_blobs(self):
             t1 = text[:_]
             self.check_serde(t1)
 
+    def test_blob_limit(self):
+        s = ' ' * 0x400000001
```
arvidn, line 48: it seems this doesn't work on CI. I imagine we will never support atoms this large anyway. Would it make sense to alter the serialization format to have a lower limit?

clvm/serialize.py
```diff
@@ -124,6 +124,8 @@ def _atom_from_stream(f, b, to_sexp):
             raise ValueError("bad encoding")
         size_blob += b
     size = int.from_bytes(size_blob, "big")
+    if size >= 0x400000000:
```
arvidn, line 127: I'm pretty sure this was a DoS vector. however, it still might be. I think we should consider lowering the max bob size

aqk: Yeah, it's pretty big.

richardkiss: For sure if it does a malloc, this is a DOS and the stream could crash. We ought to test this. I think in practice, the read is done on a `io.BytesIO` object, which might grab a substring of an existing string, so we'd just see an end-of-stream error (since the stream can't be long enough).

richardkiss: If we can verify this as a DOS vector, we might consider adding an extra parameter `max_atom_length` or something to the deserialize code. If we make it larger than the maximum generator size, the exact value is not consensus-critical.

richardkiss: (I've also considered writing the deserializer in clvm, which is possible now that we have `substr` and `len` and the bitfield operators. Then, our atoms would just be pointers into the generator bock as long substrings are shared by `substr`.)

richardkiss: Hmmm, this number is ~17 GB, so I'm not this change actually helps on machines with less than 17 GB of free memory (which is most computers).

arvidn: yeah, this is the most conservative limit. Since we have this limit in the encoder, we would reasonably not support parsing atoms this large. This is not really meant to be an attempt to address the potential DoS

richardkiss: I realize now that this limit already exists in serializing, so adding it to deserializing is about making things symmetric, not DOS mitigation.

PR #24, opened 2020-12-21
full test coverage of SExp

arvidn:


tests/as_python_test.py
```diff
@@ -10,6 +10,10 @@ class dummy_class:
     def __init__(self):
         self.i = 0
 
+def gen_tree(depth):
+    if depth == 0: return SExp.to(1337)
+    return SExp.to((gen_tree(depth - 1), gen_tree(depth - 1)))
```
richardkiss, line 15:
```suggestion
    subtree = gen_tree(depth-1)
    return SExp.to((subtree, subtree))
```
I think this will be `O(n)` instead of `O(2**n)`.

tests/as_python_test.py
```diff
@@ -10,6 +10,10 @@ class dummy_class:
     def __init__(self):
         self.i = 0
 
+def gen_tree(depth):
+    if depth == 0: return SExp.to(1337)
```
richardkiss, line 14: The `flake8` linter doesn't like this, can you put the `return` on the next line?

arvidn: fixed (along with some other flake8 complaints). Do you think we should run flake8 on CI? (and if so, what do you think our max line length should be)

richardkiss (approved):

PR #25, opened 2020-12-21
remove unused unknown-op-handler

arvidn:


richardkiss:
This patch can't be merged as we need this code for the "dialects" feature, which is implemented in `chia-blockchain` but not yet tested here. There are two dialects: one which fails on unknown opcodes, and another which returns `0` for unknown opcodes.

arvidn:
ok, great. I'll write a unit test

PR #28, opened 2020-12-21
Quote cons pair

aqk:


aqk:
https://github.com/Chia-Network/clvm_tools/pull/16

richardkiss (commented):
Did you decide to not bother with the `quote` step, which would provide a stepping-stone of migration?

tests/brun/any-3.txt
```diff
@@ -1,3 +1,3 @@
-brun -c '(all (q 0) (q 0))'
```
arvidn, line 1: the typo "all" -> "any" here has been fixed in `develop`. You should rebase to pick up this fix.

PR #31, opened 2020-12-24
Switch to Coveralls GHA

hoffmang9:
This should have a slightly more useful UI than just using the command line tool


hoffmang9:
The supported action is a mess - nevermind.

PR #32, opened 2020-12-28
support string being passed to SExp.to()

arvidn:
and have it converted to a binary blob (treated as UTF-8)

### DIFF
--- a/clvm/SExp.py
+++ b/clvm/SExp.py
@@ -61,6 +61,8 @@ class SExp(CLVMObject):
             return NULL
         if v == []:
             return NULL
+        if isinstance(v, str):
+            return v.encode()
 
         if hasattr(v, "__iter__"):
             pair: SExpType = NULL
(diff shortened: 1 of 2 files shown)


clvm/SExp.py
```diff
@@ -61,6 +61,8 @@ def _to_sexp_type(
             return NULL
         if v == []:
             return NULL
+        if isinstance(v, str):
```
aqk, line 64: This looks really handy. Two small things I worry about with this:

* Mechanism / comment to warn about using this from consensus/blockchain code (if someone adds a CLVM operator in the future, warn them away from this, because it goes through Python's string conversion)
* UTF-8 conversion surprise

aqk (approved):

clvm/SExp.py
richardkiss: I can't decide if we should quietly support strings or if we should raise an error. The downside of supporting strings is it lets users be sloppy with `bytes` vs `str`. The upside is it lets users be sloppy with `bytes` vs `str`. Maybe that's okay, since the main use case for `to` is for use with quick-hack slop code.

PR #33, opened 2020-12-28
make SExp.to() non-recursive

arvidn:
to not fail because of stack overflow. I also extended the tests with large structures (that otherwise would cause a stack overflow).

The reason these new tests can't use `self.check_as_python()` is because it uses the built-in comparison function in python, which is recursive, and fails with stack overflow.

I also tried to add this test:

```
self.check_as_python(((b"2", b"3"), (b"", b"1")))
```

But it failed by producing the wrong output. It does in current `develop` as well, so it seems the `SExp.to()` and `as_python()` don't round-trip. @richardkiss is that a problem?

### DIFF
--- a/clvm/SExp.py
+++ b/clvm/SExp.py
@@ -39,41 +39,77 @@ class SExp(CLVMObject):
         class_,
         v: CastableType,
     ) -> SExpType:
-        if isinstance(v, tuple):
-            if len(v) != 2:
-                raise ValueError("can't cast tuple of size %d" % len(v))
-            left, right = v
-            if type(left) != CLVMObject:
-                left = CLVMObject(class_._to_sexp_type(left))
-            if type(right) != CLVMObject:
-                right = CLVMObject(class_._to_sexp_type(right))
-            return (left, right)
-        if isinstance(v, CLVMObject):
-            return v.pair or v.atom
-        if isinstance(v, bytes):
-            return v
+        stack = [v]
+        ops = [(0, None)] # convert
 
-        if isinstance(v, int):
-            return int_to_bytes(v)
-        if isinstance(v, G1Element):
-            return bytes(v)
-        if v is None:
-            return NULL
-        if v == []:
-            return NULL
-        if isinstance(v, str):
-            return v.encode()
+        while len(ops) > 0:
+            op, target = ops.pop()
+            # convert value
+            if op == 0:
+                v = stack.pop()
+                if isinstance(v, tuple):
+                    if len(v) != 2:
+                        raise ValueError("can't cast tuple of size %d" % len(v))
+                    left, right = v
+                    target = len(stack)
+                    stack.append((left, right))
+                    if type(right) != CLVMObject:
+                        stack.append(right)
+                        ops.append((2, target)) # set right
+                        ops.append((0, None)) # convert
+                    if type(left) != CLVMObject:
+                        stack.append(left)
+                        ops.append((1, target)) # set left
+                        ops.append((0, None)) # convert
+                    continue
+                if isinstance(v, CLVMObject):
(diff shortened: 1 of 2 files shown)


arvidn:
It's a bit of a mystery to me why my patch is so slow. I'm not super experienced with the performance of various python constructs, I would just have thought that `append()` and `pop()` on a list would be very quick, compared to recursive function calls.

What I've got so far are these comparisons of profiles. Specifically, it's the serialization test `deep_tree` that's taking forever. It doesn't hand indefinitely, it just looks like it's order square or cube, and perhaps some constant increased enough to just make the large test cases explode (time wise).

One observation is that we call `as_pair()` a lot, which is not a cheap operation. It's O(n) for nested structures. I haven't quite figured out what the internal format of `CLVMObject` and `SExp` is, but I would have expected there to be a way to ask for the pair without having to convert the entire tree.

This is the profile of *this* patch:
deep_tree

This is the same profile of tip of the *develop* branch:
deep_tree-develop

richardkiss:
> I also tried to add this test:
> 
> ```
> self.check_as_python(((b"2", b"3"), (b"", b"1")))
> ```
> 
> But it failed by producing the wrong output. It does in current `develop` as well, so it seems the `SExp.to()` and `as_python()` don't round-trip. @richardkiss is that a problem?

This is okay, as `((2 . 3) . (0 . 1))` and `((2 . 3) 0 . 1)` are the same (and so python `((2, 3), (0, 1))` is the same as `([(2, 3), 0] . 1)`.

richardkiss:
I see this slowdown problem in the rust implementation too, which makes me think there is some recursion happening somewhere where it shouldn't be. The theory is that an `SExp` wraps either a blob or a 2-tuple of `(c1, c2)` where `c1` and `c2` are type `CLVMObject`. When you call `as_pair`, it returns a 2-tuple of `(s1, s2)` where `s_i` is of type `SExp`, each of which wraps the `CLVMObject` type of `c_i`. Something is going wrong somewhere in `as_pair` where it incorrectly dips down the whole tree.

arvidn:
I rebased to incorporate the patch to accept `str` as well as added a fix (the top commit) to fix the performance regression

clvm/SExp.py
```diff
@@ -98,13 +98,13 @@ def _to_sexp_type(
 
                 raise ValueError("can't cast to %s: %s" % (class_, v))
             if op == 1: # set left
-                stack[target] = (stack.pop(), stack[target][1])
+                stack[target] = (CLVMObject(stack.pop()), stack[target][1])
```
aqk, line 101: Are these conversions because the things coming off the stack might be Python tuples?

arvidn: the short answer is that this fix restores the original behavior. I understood this code a bit better after richard explained it last night; the conversions in `as_tuple()` aren't supposed to be deep, they are supposed to be short-circuited by the types already being a `CLVMObject` (or `SExp`) type. This makes sure they actually are. Otherwise the whole tree has to be traversed to wrap everything every time `as_tuple()` is called.

clvm/SExp.py
```diff
@@ -39,41 +39,77 @@ def _to_sexp_type(
...
+                    continue
+
+                if hasattr(v, "__iter__"):
+                    target = len(stack)
+                    stack.append(NULL)
+                    for _ in v:
+                        stack.append(_)
+                        ops.append((3, target)) # prepend list
+                        # we only need to convert if it's not already the right
+                        # tpye
```
richardkiss, line 94: typo

richardkiss: or should I say tpyo

richardkiss (approved):

PR #35, opened 2021-01-04
add div operator

arvidn:
the output from `div-02.txt` seems a bit strange to me:

```
FAIL: div requires 2 args (q)
```


arvidn:
oh, it's because `1` turns into `q`...

clvm/more_ops.py
```diff
@@ -147,6 +149,16 @@ def op_divmod(args):
     return cost, args.to((q, r))
 
 
+def op_div(args):
+    cost = DIV_BASE_COST
+    i0, i1 = args_as_int_list("div", args, 2)
```
aqk, line 154: This looks good. Is the performance much different than divmod?

I guess we need to wait for Rust benchmarking

aqk (commented):
@bramcohen @richardkiss This div without remainder. I assume we will benchmark this with the Rust version, and include it if it is much faster than div with remainer.

IS that the plan?

richardkiss:
Copying my comment from a conversation in https://github.com/Chia-Network/clvm/pull/16 ...

```
(venv) kiss@oasis develop % python -m timeit -s "T = 10004843716274856703848" "T >> 6"
5000000 loops, best of 5: 59.1 nsec per loop
(venv) kiss@oasis develop % python -m timeit -s "T = 10004843716274856703848" "T // 64"
5000000 loops, best of 5: 87.1 nsec per loop
(venv) kiss@oasis develop % python -m timeit -s "T = 100" "T // 64" 
10000000 loops, best of 5: 34.6 nsec per loop
(venv) kiss@oasis develop % python -m timeit -s "T = 100" "T >> 6" 
5000000 loops, best of 5: 59.1 nsec per loop
```

Our values on the left are likely to be closer to 100 than 10004843716274856703848, so turns out that `//` is within order of magnitude even in the worst case. My intuitions were off here.

richardkiss:
Maybe this example uses too small a value for `A` and `B`. Let me try again with bigger values.

richardkiss:
```
(clvm) kiss@oasis clvm % A=481095061974804308368284666404899763637029381567209856270837; B=68479775184784685330020319572759593676822108799802
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "A//B"
2000000 loops, best of 5: 168 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "A%B" 
2000000 loops, best of 5: 167 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "(A//B, A%B)" 
1000000 loops, best of 5: 371 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "divmod(A,B)" 
1000000 loops, best of 5: 231 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "Q = A//B; (Q, A - B * Q)"
1000000 loops, best of 5: 328 nsec per loop
```

Smaller B value

```
(clvm) kiss@oasis clvm % A=481095061974804308368284666404899763637029381567209856270837; B=684           
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "A//B"
2000000 loops, best of 5: 156 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "A%B"                                  
2000000 loops, best of 5: 157 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "(A//B, A%B)"                          
1000000 loops, best of 5: 308 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "divmod(A,B)"                          
1000000 loops, best of 5: 207 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "Q = A//B; (Q, A - B * Q)"             
1000000 loops, best of 5: 293 nsec per loop
```

richardkiss:
```(clvm) kiss@oasis clvm % A=74738383914845; B=83717484                                         
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "A//B"                                 
2000000 loops, best of 5: 104 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "A%B"                     
2000000 loops, best of 5: 103 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "(A//B, A%B)"             
1000000 loops, best of 5: 205 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "divmod(A,B)"             
2000000 loops, best of 5: 158 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "Q = A//B; (Q, A - B * Q)"
1000000 loops, best of 5: 209 nsec per loop
(clvm) kiss@oasis clvm % python -m timeit "A=$A; B=$B" "(A, B)"                  
5000000 loops, best of 5: 46.7 nsec per loop
```

It looks like nearly all the overhead of `divmod` vs `//` or `%` alone is due to the tuple being created! For large A and B, `divmod` is about 50% than calculating the remainder yourself. All in all, it's pretty inconclusive about what to do. So maybe we add `//` and keep divmod.

richardkiss (approved):

arvidn:
I benchmarked clvm programs with `div` and `divmod`, and they came out to about the same. The largest numbers I used in those tests were 1024 bytes long. However, on this same machine as I ran all the benchmarks on before, `divmod` took almost twice as long as last time, so just including the `div` result wouldn't be reasonable. Instead I used the same cost for `div` as it currently measures to about the same as `divmod`.

So there is still a bit of a mystery why my benchmarks no longer reproduce.

PR #36, opened 2021-01-05
run flake8 on CI

arvidn:


arvidn:
this needs to be rebased on top of https://github.com/Chia-Network/clvm/pull/24 once that lands. Then I will drop the temporary commit (which currently was cherry-picked from that PR)

clvm/operators.py
```diff
@@ -50,7 +50,8 @@ def __new__(class_, d: Dict):
     def __call__(self, op: bytes, arguments: CLVMObject) -> Tuple[int, CLVMObject]:
         f = self.get(op)
         if f is None:
-            f = lambda args: self.unknown_op_handler(op, args)
+            def f(args):
+                return self.unknown_op_handler(op, args)
```
arvidn, line 54: apparently default flake8 prefers this. I'm trying hard not to judge :)

richardkiss: I guess `lambda` requires understanding of functional programming style, which is a tad meta.

richardkiss (approved):

PR #38, opened 2021-01-06
Fix problem with `SExp.to([1])`.

richardkiss:


arvidn:
what was the problem? (I suppose I still don't fully understand what the intended types for the tree are). It looks like everything is supposed to be CLVMObject now. But the code prior to my patch had the behavior of using `class_` in the list case, I'm pretty sure.

It seems to me it may be possible to merge `SExp` and `CLVMObject`. The former is the only class deriving from `CLVMObject` and `CLVMObject` isn't used other than through `SExp`. Or perhaps I'm missing some subtlety.

arvidn:
I really think we need a test that covers the problem this fixes.

arvidn:
This test pass in current `develop`:

```
diff --git a/tests/as_python_test.py b/tests/as_python_test.py
index 1000691..d9f67ee 100644
--- a/tests/as_python_test.py
+++ b/tests/as_python_test.py
@@ -49,6 +49,11 @@ class AsPythonTest(unittest.TestCase):
         v = SExp.to([])
         self.assertEqual(v.atom, b"")
 
+    def test_list_of_one(self):
+        v = SExp.to([1])
+        self.assertEqual(v.pair[0], b"\x01")
+        self.assertEqual(v.pair[1], b"")
+
     def test_g1element(self):
         b = b"\xb3\xb8\xac\x53\x7f\x4f\xd6\xbd\xe9\xb2" \
             b"\x62\x21\xd4\x9b\x54\xb1\x7a\x50\x6b\xe1\x47\x34\x7d\xae\x5d" \
```

This should probably be incorporated

arvidn (approved):
looks good, I think you should include my patch to this PR before landing it though

arvidn (approved):

richardkiss:
> It seems to me it may be possible to merge `SExp` and `CLVMObject`. The former is the only class deriving from `CLVMObject` and `CLVMObject` isn't used other than through `SExp`. Or perhaps I'm missing some subtlety.

The rust implementation includes a replacement for `CLVMObject`, but there's not much gain in porting all the `SExp` functionality to rust.

PR #41, opened 2021-01-08
Various fixes

richardkiss:
Each change is a separate, disjoint fix.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.7', '3.8' ]
+        python-version: [ '3.7', '3.8', '3.9' ]
     name: Python ${{ matrix.python-version }} sample
     steps:
       - uses: actions/checkout@v2
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This is the in-development version of a LISP-like language for encumbering and releasing funds with smart-contract capabilities.
 
-See docs/clvm.org for more info.
+See docs/clvm.org or https://chialisp.com/ for more info.
 
 
 Testing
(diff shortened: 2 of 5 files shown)


clvm/run_program.py
```diff
@@ -54,8 +56,12 @@ def run_program(
         pre_eval_op = None
 
     def traverse_path(sexp: SExp, env: SExp) -> Tuple[int, SExp]:
-        cost = PATH_LOOKUP_COST_PER_LEG
-        if sexp.nullp():
+        cost = TRAVERSE_BASE_COST
+        atom = sexp.atom
+        while len(atom) and atom[0] == b"\0":
```
arvidn, line 61: I think it would be easier to understand this if the cost was called `TRAVERSE_COST_PER_REDUNDANT_BYTE` or something to indicate that it's a counter-measure.

But wouldn't it be better to just not allow any redundant bytes in the blob? i.e. just hard-fail if there are any bytes left at the end.

clvm/run_program.py
```diff
@@ -54,8 +56,12 @@ def run_program(
...
+            cost += TRAVERSE_COST_PER_BYTE
+            atom = atom[1:]
+        if len(atom) == 0:
```
arvidn, line 64: shouldn't this be a failure? The most significant bit, iirc, terminates the traversal. If you don't have a most significant bit, it seems reasonable to treat that as an error.

arvidn:
Everything looks good to me except the new `TRAVERSE_COST_PER_BYTE` logic. Is there a good reason to not hard-fail on "overlong" paths instead?

clvm/run_program.py
richardkiss: The current behaviour is to treat the number as an integer, which may actually require the leading 0x00. I don't think we have many tests of deep traversal, so we probably need them to find edge cases and test costs.

clvm/run_program.py
richardkiss: Intuitively, I think of a string of 0 bytes as a non-minimal 0, which has a specific meaning (evaluates to nil). I'm not sure what the benefit of failure is. Maybe we have a strict mode for debugging and non-strict mode for consensus?

clvm/run_program.py
arvidn: in the performance benchmark I generate tests of deep lookups, but they aren't testing correctness

clvm/run_program.py
arvidn: > Intuitively, I think of a string of 0 bytes as a non-minimal 0, which has a specific meaning (evaluates to nil).

0 is not a valid path though, right?

I would think the benefit of failure are:

* it's a grey zone that's not necessary and likely to not be intuitive. Sort of like UTF-8 overlong representations being illegal.
* it's easier to make it well-formed if we come up with a reasonable meaning in the future, since no valid program will rely on it. I suppose we can't rely on all nodes accepting a new behavior though.

clvm/run_program.py
arvidn: I agree that regardless of the validity of these paths, we need new tests to cover this

arvidn: > The current behaviour is to treat the number as an integer, which may actually require the leading 0x00.

We only treat it as an integer to the extent that we convert it to a bigint value. iirc, we don't really care whether the value is negative or positive, we just look at one bit at a time, right? It's not clear to me that there *are* any cases were you need to add a 0-prefix. However, I realize the code might look a bit funny having negative paths in it. Although, we could have a convention to represent paths as hex blobs.

clvm/run_program.py
richardkiss: `0` actually is valid... it used to be the same as `1`, but within the past few weeks, it's changed to evaluate to `0`, as an analog to the 0 register in some RISC CPUs. The compiler does not yet use this optimization.

richardkiss: Also, I'm pretty sure the compiler currently uses non-minimal paths, which isn't really ideal.

richardkiss:
I've removed the controversial changes and force-pushed, so we can merge the non-controversial ones. @arvidn

arvidn (approved):

PR #44, opened 2021-01-13
make to_sexp_type a free function

arvidn:
it doesn't use anything from the SExp class so it doesn't need to be a member

### DIFF
--- a/clvm/SExp.py
+++ b/clvm/SExp.py
@@ -29,87 +29,86 @@ CastableType = typing.Union[
 NULL = b""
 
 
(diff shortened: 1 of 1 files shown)


richardkiss:
~~The `Program` class in `chia-blockchain` actually subclasses `SExp`, so this change would break that.~~ Never mind, we just need `to` to be included, so this is a good change. Can you rebase? And I'll merge it.

arvidn:
rebased

richardkiss (approved):

PR #45, opened 2021-01-13
Fix problem with `sexp.to([1, sexp])` under `clvm_rs`.

richardkiss:


arvidn (approved):
you don't think we should have a test, to avoid a regression?

richardkiss:
The test in `to_sexp_test.py` fails without this change (although only if `clvm_rs` is active).

PR #47, opened 2021-01-14
Cost leading zeroes

arvidn:
inspired by the rust implementation; instead of counting "limbs" after parsing the blob into an integer, charge cost per *byte*. i.e. leading zeroes still cost as if they were not zero.

I think this is a simple and cheap way of charging for leading zeroes.


clvm/more_ops.py
```diff
@@ -65,7 +65,7 @@ def args_as_ints(op_name, args):
     for arg in args.as_iter():
         if arg.pair:
             raise EvalError("%s requires int args" % op_name, args)
-        yield arg.as_int()
+        yield (arg.as_int(), len(arg.as_atom()))
```
arvidn, line 68: does this make sense?
I realized I should do this for `args_as_bools()` too

richardkiss (approved):

arvidn:
There are a few things left for this, I would like your opinions @aqk @richardkiss :

`op_substr` does not charge for leading zeroes of the indices
`op_ash` and `op_lsh` don't charge for leading zeroes of the second parameter, the steps to shift.
`op_not` does not charge for leading zeroes, but this seems OK, since all we do is to check if it's `b""`.

I posted this followup, but it doesn't address these questions: https://github.com/Chia-Network/clvm/pull/52

PR #48, opened 2021-01-14
unify pattern for error messages

arvidn:
when passing a list where an atom is expected
Consistently print all arguments passed to the operator in the error message


tests/brun/concat-4.txt
```diff
@@ -1,2 +1,2 @@
 brun -c '(concat (q foo) (q (bar)))'
-FAIL: concat on list ("bar")
+FAIL: concat on list ("foo" ("bar"))
```
aqk, line 2: These lists aren't being deserialized correctly.

arvidn: hm. both rust and python prints them like this, as if it's deliberate. @richardkiss ?

richardkiss: I don't understand the objection. The arguments to `concat` are exactly `("foo" ("bar"))` in this example.

richardkiss:
At the risk of splitting hairs off a bikeshed, doesn't it seem to make more sense on `XXX on list` to print *just* the single argument that caused the problem, rather than enumerating all arguments? There are arguments each way. Pro: it's more clear which argument causes the problem. Con: it might be more clear (especially if there are more than one argument) which invocation of the operator in your program caused the problem. What do you think @aqk @arvidn ?

arvidn:
I'm in favor of just printing the argument that caused the problem. From what I can tell, we have two forms of this error:

1. "XXX requires int args"
2. "XXX on list"

They are basically the same kind of error and initially I considered unifying them as well, but I thought maybe that would be going too far.

In (1), we always print all the args (except "softfork"), in some cases in (2) we print just the non-atom argument. This PR is the smallest change to unify it, but going the other way would simplify the code further, as far as I can tell.

arvidn:
I've updated (force pushed) this patch to only print the argument that failed the check (i.e. wasn't an atom). The patch is larger but I think it's better.

clvm/core_ops.py
```diff
@@ -54,7 +54,7 @@ def op_eq(args):
     a0 = args.first()
     a1 = args.rest().first()
     if a0.pair or a1.pair:
-        raise EvalError("= on list", args)
+        raise EvalError("= on list", a0)
```
richardkiss, line 57: Aha! Here's why `args` was used: it might be `a0` is an atom, and `a1` is the list.

clvm/more_ops.py
```diff
@@ -171,7 +171,7 @@ def op_gr_bytes(args):
         raise EvalError(">s takes exactly 2 arguments", args)
     a0, a1 = arg_list
     if a0.pair or a1.pair:
-        raise EvalError(">s on list", args)
+        raise EvalError(">s on list", a0)
```
richardkiss, line 174: Same issue here: the problem might be `a1`.

clvm/core_ops.py
richardkiss: Possible solution:

`raise EvalError("= on list", a0 if a0.pair else a1)`

arvidn:
updated. I also made sure to add unit tests for these cases that weren't covered

arvidn:
and rebased

richardkiss (approved):

aqk:
About printing the arguments: I mainly use the error message to locate the expression with the error, more so than figuring out what the error was, so if e.g. `(i 0 4)` gives "too few args", it's a bit nicer to have the full expression.

The fanciest is to print the sub-expression that caused the issue, and highlight the problem, but we don't need that right now.

Either way is good for now.

PR #49, opened 2021-01-14
fail early when pulling out arguments as lists

arvidn:


arvidn:
perhaps I'm a bit paranoid. I don't have a every strong opinion about this, and the `bool` counterpart is still unchanged, since we charge cost based on number of arguments, not the number of arguments we "visit", so we need to pull them all out up-front in the bool case.

richardkiss:
I'm not sure this patch is necessary, as it can only really be a DOS vector for a large number of integers. But no transaction that raises an `EvalError` can ever appear in a block, or even in a mempool. If it didn't have that repeated `raise` line, it'd be a no-brainer... as it is, I'm not convinced. What do you think?

arvidn:
yeah, I agree. If there's an "infinite" argument list, the parser would barf before this code

PR #50, opened 2021-01-14
Rename QUOTE_KW to QUOTE_ATOM

aqk:


clvm/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .SExp import SExp
 from .operators import (  # noqa
+    QUOTE_ATOM,
```
arvidn, line 3: exposing my python-ignorance here; Does this import, being in `__init__` automatically cause it to be imported in all python files in this directory? If not, I don't understand the purpose of this.

clvm/operators.py
```diff
@@ -65,3 +64,4 @@ def unknown_op_raise(self, op: bytes, arguments: CLVMObject):
     operators_for_module(KEYWORD_TO_ATOM, core_ops, OP_REWRITE)
 )
 OPERATOR_LOOKUP.update(operators_for_module(KEYWORD_TO_ATOM, more_ops, OP_REWRITE))
+QUOTE_ATOM = KEYWORD_TO_ATOM["q"]
```
arvidn, line 67: My understanding is that `OPERATOR_LOOKUP` is "exported" and can be used when calling `run_program()` to get the normal operators. But `run_program()` can really be customized to take another set of operators with different names. How does the `QUOTE_ATOM` fit in to this? is that also a customization point, or is there an implicit requirement now that any custom operator list must have its quote operator be "1"?

clvm/__init__.py
aqk: No, this is so you can write:

`from clvm import QUOTE_ATOM`

in other modules that import clvm. It's just exposing that binding at the root of the module, instead of having to say:

`from clvm.operators import QUOTE_ATOM`

clvm/operators.py
aqk: TL;DR: The intent was that using `QUOTE_ATOM` would mean changing `q` to `quote` in only one place, and 

> is there an implicit requirement now that any custom operator list must have its quote operator be "1"
No, as long as `quote_kw` is passed into `run_program`, and the bytecode was compiled with that alternate `q` in mind, that invocation of `run_program` will work.

BTW, the reason behind what I'm doing here is, I wanted to to export the *text* representation of the quote keyword from this module ("q") here, and use that in other places in the compiler that required us to recognise the quote symbol as text - eg. the tokenizer. I thought I needed that so we could execute some staged tests that Richard suggested - (eg. compiling `quote` to `1`, and turning off `q` in the tokenizer).

But in doing that, I realized that QUOTE_KW was a misnomer, so I did this smaller patch first, intending to export both QUOTE_KW and QUOTE_ATOM in a subsequent patch.

The reason to not use OPERATOR_LOOKUP is that the lookup depends on the string subscript to the dict. Using QUOTE_ATOM removes the possibility of the mistake of using `KEYWORD_TO_ATOM["q"]` when we meant to use `b"quote"`, while both `quote` and `q` coexist in the tokenizer.

aqk: See https://github.com/Chia-Network/clvm_tools/pull/22

arvidn (approved):

PR #51, opened 2021-01-15
Quote1

aqk:


aqk:
Note that this won't pass tests without https://github.com/Chia-Network/clvm_tools/pull/20

clvm/operators.py
```diff
@@ -8,8 +8,9 @@
 from .casts import int_to_bytes
 from .op_utils import operators_for_module
 
+QUOTE_KW = "quote"
```
arvidn, line 11: we're changing it from `q` to `quote`?

arvidn: I assume this is to ensure there are no `q`s accidentally lurking anywhere that we forgot to change. The down-side with this is that it makes this patch touch everything, and it makes it harder to interleave into the commit-stream

aqk: Yes, I think this was the reasoning behind Richard asking for this change.

aqk: Just talked with Richard. Please disregard this PR until the next update. Thanks!

PR #54, opened 2021-01-26
additional tests for cost of path-lookups

arvidn:


tests/brun/path-4.txt
```diff
@@ -0,0 +1,3 @@
+brun -n -c '1' '(((0x1337 . (0x1337 . (42 . 0x1337))) . 0x1337) . 0x1337)'
+cost = 53
+(((4919 4919 42 . 4919) . 4919) . 4919)
```
arvidn, line 3: You may want to note that we print this in a (possibly) unintuitive way. I was thrown off at least. I believe it's consistent though.

tests/brun/path-5.txt
```diff
@@ -0,0 +1,3 @@
+brun -n -c '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001' '(((0x1337 . (0x1337 . (42 . 0x1337))) . 0x1337) . 0x1337)'
+cost = 53
```
arvidn, line 2: @richardkiss You may notice that leading zeros don't seem to cost anything here.

tests/brun/path-4.txt
aqk: Ya, I know what you mean. To some extent, it can't be helped, given the syntax that lisp chose.

tests/brun/path-3.txt
```diff
@@ -0,0 +1,3 @@
... ((0x1337 . (((0x1337 . (0x1337 . (0x1337 . ((0x1337 . (((0x1337 . (((0x1337 . (0x1337 . (0x1337 . (0x1337 . ((0x1337 . ((0x1337 . (((((0x1337 . ((0x1337 . ((0x1337 . (0x1337 . (0x1337 . (((0x1337 . (0x1337 . ((0x1337 . (0x1337 . ((((0x1337 . (0x1337 . (0x1337 . (0x1337 . (((((0x1337 . (0x1337 . (0x1337 . (0x1337 . (0x1337 . (((((0x1337 . (((((0x1337 . ((0x1337 . (0x1337 . ((((0x1337 . ((((0x1337 . ((0x1337 . ((0x1337 . ((0x1337 . (0x1337 . (0x1337 . ((((0x1337 . (0x1337 . ((0x1337 . (((0x1337 . (0x1337 . (((0x1337 . (0x1337 . (0x1337 . (42 . 0x1337)))) . 0x1337) . 0x1337))) . 0x1337) . 0x1337)) . 0x1337))) . 0x1337) . 0x1337) . 0x1337)))) . 0x1337)) . 0x1337)) . 0x1337)) . 0x1337) . 0x1337) . 0x1337)) . 0x1337) . 0x1337) . 0x1337))) . 0x1337)) . 0x1337) . 0x1337) . 0x1337) . 0x1337)) . 0x1337) . 0x1337) . 0x1337) . 0x1337)))))) . 0x1337) . 0x1337) . 0x1337) . 0x1337))))) . 0x1337) . 0x1337) . 0x1337))) . 0x1337))) . 0x1337) . 0x1337)))) . 0x1337)) . 0x1337)) . 0x1337) . 0x1337) . 0x1337) . 0x1337)) . 0x1337)) . 0x1337))))) . 0x1337) . 0x1337)) . 0x1337) . 0x1337)) . 0x1337)))) . 0x1337) . 0x1337)) . 0x1337))) . 0x1337)))))) . 0x1337)) . 0x1337) . 0x1337))) . 0x1337)))) . 0x1337))'
```
aqk, line 1: Eyeballing this test, that tree doesn't look deep enough for that huge path specifier. What's going on there?

arvidn: it's effectively a list, so the depth of the tree is the same as the number of `0x1337` that you see. So, the depth is more than 64, so the path requires a number greater than 2^64.

tests/brun/path-5.txt
richardkiss: Hmm, they ought to cost a minimal amount, like 1. I thought I'd implemented that.

tests/brun/path-3.txt
arvidn: i.e. this is testing that paths > `INT64_MAX` are supported

richardkiss (approved):

tests/brun/path-5.txt
richardkiss: I see, I think I implemented an additional cost of leading zeros for rust, but not for python. I prefer how rust does it. We need to unify this algorithm. This test will help catch the discrepancy.

richardkiss:
Can we add a test that uses bitstreams with the high bit set? Looks like there is an error in the python implementation, since it converts to an int. 0xff = -1 seems to cause problems (it treats it like 1 instead of 0xff). Unifying with the rust implementation would fix this problem.

arvidn:
by "bitstream", do you mean the path? Sure, I can add that test.
These tests fail with the rust implementation. I'm working on fixing that

PR #55, opened 2021-01-26
Path test

arvidn:
This implements the path traversal more efficiently by not shifting a bigint for every step.
Perhaps more importantly, it doesn't interpret the bytes as a signed integer in the first place, fixing a mis-interpretation of paths like `0xff` (turning into -1).

In fact, the new path traversal simply interprets the path as a blob of bits.

This patch also adds a cost to (redundant) leading zeros.

Interestingly, such redundant leading zeroes are added sometimes implicitly when paths are specified as integers (rather than hex blobs). Specifically, if the most significant bit is 1, a 0-padding byte is prepended. This is redundant and now comes with a cost.


tests/brun/path-6.txt
```diff
@@ -1,3 +1,3 @@
 brun -n -c '56800' '((((((0x1337 . (0x1337 . (0x1337 . (0x1337 . ((0x1337 . (0x1337 . (0x1337 . ((0x1337 . 42) . 0x1337)))) . 0x1337))))) . 0x1337) . 0x1337) . 0x1337) . 0x1337) . 0x1337)'
-cost = 68
+cost = 69
```
arvidn, line 2: it might be unintuitive at first, why these costs changed. However, keep in mind that since the path is entered as a positive decimal number, and the most significant bit is set, we implicitly add a leading zero to ensure it's not interpreted as a negative number. It would be more efficient to specify this path as `0xDDE0`, since it would not have a leading zero in that case.

richardkiss: You might consider adding a test that does just that (`0xdde0` instead of `56800`) .

clvm/run_program.py
```diff
@@ -57,16 +57,39 @@ def traverse_path(sexp: SExp, env: SExp) -> Tuple[int, SExp]:
...
+        end_byte_cursor = 0
+        while end_byte_cursor < len(b) and b[end_byte_cursor] == 0:
+            end_byte_cursor += 1
+
+        cost += end_byte_cursor * PATH_LOOKUP_COST_PER_ZERO_BYTE
+        if end_byte_cursor == len(b):
+            return cost, sexp.null()
+
+        # create a bitmask for the most significant *set* bit
+        # in the last non-zero byte
```
richardkiss, line 72: Wow, how does this work? Is this from https://graphics.stanford.edu/~seander/bithacks.html ?

richardkiss:
Looks great! If you think we should add that extra test, go ahead & do that. If not, also let me know, and I'll merge.

tests/brun/path-6.txt
arvidn: there are tests for that too. path-9 and path-10 iirc

clvm/run_program.py
arvidn: yes, inspired by: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2

tests/brun/path-6.txt
arvidn: nooo. I must have forgotten `git add path-10.txt` :/

clvm/run_program.py
arvidn: there really should be a unit test for this, since it's not obviously correct by inspection

richardkiss (approved):

clvm/run_program.py
```diff
@@ -57,16 +64,35 @@ def traverse_path(sexp: SExp, env: SExp) -> Tuple[int, SExp]:
...
```
aqk, line 79: When you say "last non-zero byte", it might be more clear to say "least significant non-zero byte"

arvidn: yeah, I was thinking about this when I was reviewing this code. When I say "last" here, I mean the from the right, in a big-endian kind-of-way. It's the *first* non-zero byte in the buffer, because we're skipping leading zeroes. I made this mistake in my first revision, thinking about the number as little-endian

PR #57, opened 2021-01-28
Quote transform 2

aqk:


tests/brun/add-1.txt
```diff
@@ -1,3 +1,3 @@
-brun -c '(+ (q 7) (q 1))'
+brun -c '(+ (q . 7) (q . 1))'
```
arvidn, line 1: do all clvm programs need this syntax once this change lands?

arvidn (approved):

PR #59, opened 2021-02-01
consume serialized SExp objects, without constructing the python tree

arvidn:
This is a cheaper version of "parsing" serialized S-expressions. Instead of parsing it into a python tree of CLVM objects, it consumes the bytes from the stream that represent one S-Expression tree, and returns those bytes. This is a way to preserve the serialized form of a CLVM program.

This function is intended to be used by Chia Blockchain to preserve serialized programs, that can be passed into the rust implementation and parsed in rust.

### DIFF
--- a/tests/serialize_test.py
+++ b/tests/serialize_test.py
@@ -2,7 +2,7 @@ import io
 import unittest
 
 from clvm import to_sexp_f
-from clvm.serialize import (sexp_from_stream, atom_to_byte_iterator)
+from clvm.serialize import (sexp_from_stream, sexp_buffer_from_stream, atom_to_byte_iterator)
 
 
 TEXT = b"the quick brown fox jumps over the lazy dogs"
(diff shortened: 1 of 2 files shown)


tests/serialize_test.py
```diff
@@ -38,6 +38,13 @@ def check_serde(self, s):
             b = v.as_bin()
             v1 = sexp_from_stream(io.BytesIO(b), to_sexp_f)
         self.assertEqual(v, v1)
+        # this copies the bytes that represent a single s-expression, just to
+        # know where the message ends. It doesn't build a python representaion
+        # of it
+        print(b)
```
richardkiss, line 44: Probably want to remove the `print` at some point.

arvidn: oops. left-over debugging..

richardkiss (approved):

PR #60, opened 2021-02-03
default to lenient-mode

arvidn:
where unknown opcodes are no-ops or fatal errors with varying costs


clvm/operators.py
```diff
@@ -43,22 +55,19 @@ class OperatorDict(dict):
...
         self = super(OperatorDict, class_).__new__(class_, d)
-        self.unknown_op_handler = self.unknown_op_raise
+        self.unknown_op_handler = default_unknown_op
         return self
 
     def __call__(self, op: bytes, arguments: CLVMObject) -> Tuple[int, CLVMObject]:
         f = self.get(op)
         if f is None:
-            def f(args):
-                return self.unknown_op_handler(op, args)
```
arvidn, line 53: I thought this code wasn't as straight-forward as it could be.

tests/operators_test.py
```diff
@@ -17,7 +17,7 @@ def unknown_handler(self, name, args):
         return 42, SExp.to(b'foobar')
 
     def test_unknown_op(self):
-        self.assertRaises(EvalError, lambda: OPERATOR_LOOKUP(b'unknown-op', SExp.to(1337)))
+        self.assertRaises(EvalError, lambda: OPERATOR_LOOKUP(b'1337', SExp.to(1337)))
```
arvidn, line 20: heh, I meant to change this to be an odd number, but it's still an ascii "7", not a byte 7. Still appears to be an odd number though

clvm/operators.py
aqk: +1

clvm/operators.py
```diff
@@ -35,6 +36,17 @@
 }
 
 
+def default_unknown_op(op: bytes, arguments: CLVMObject) -> Tuple[int, CLVMObject]:
```
aqk, line 39: What's the motivation for this implementation? (Even / Odd)

arvidn: My understanding is that it will enable us, to some extent, extend the instruction set in the future, and still have old clients have a reasonable alternative behavior and cost.

If we want to have a new instruction that is a hard failure for clients that don't understand it, half the opcode space give you that. If we want a new instruction that is a no-op for clients that don't understand it, that's the other half of the op-code space.

If we want the no-op to have a certain cost (within some limits), that opcode is available too

richardkiss: We might want some opcodes that have a cost that depends upon the inputs in some way that somewhat mirrors how complicated costs work in other opcodes.

richardkiss: I also think it might make sense to move this logic to `chia-blockchain` (ie. allow the long tails of the dialects to be defined there).

clvm/operators.py
```diff
@@ -43,22 +55,19 @@ class OperatorDict(dict):
...
 
     def __call__(self, op: bytes, arguments: CLVMObject) -> Tuple[int, CLVMObject]:
         f = self.get(op)
         if f is None:
-            def f(args):
-                return self.unknown_op_handler(op, args)
-        return f(arguments)
+            return self.unknown_op_handler(op, arguments)
+        else:
+            return f(arguments)
```
richardkiss, line 66: (You don't really need the `else:` here. It's a taste thing, but my taste is to not have it since it consumes space, possibly adds an opcode to the python, and isn't particularly confusing.)

clvm/operators.py
richardkiss: That does make testing with `brun` more challenging though. I wonder if `brun` should optionally take a `-s`-like function that defines a full path to a class that completely defines the dialect.

richardkiss: Along these lines, there's an argument to be made that everything in `more_ops` could actually belong in `chia-blockchain` (although it gets trickier when rust is involved), and then we move the corresponding `brun` tests there. It seems to make more sense when `clvm` core is relatively stable.

clvm/operators.py
arvidn: yeah, I'm going back-and-forth. This way one can pretend that `if`s are expressions :)

clvm/operators.py
```diff
@@ -35,6 +49,97 @@
...
+# 1: computed like operator add, multiplied by (multiplier + 1)
+# 2: computed like operator mul, multiplied by (multiplier + 1)
+# 3: computed like operator concat, multiplied by (multiplier + 1)
+
+# this means that unknown ops where fun is 1, 2, or 3, may still be fatal errors
+# if the arguments passed are not atoms.
+
+def default_unknown_op(op: bytes, args: CLVMObject) -> Tuple[int, CLVMObject]:
+    # any opcode starting with ffff is reserved (i.e. fatal error)
+    if len(op) >= 2 and op[:2] == b"\xff\xff":
```
richardkiss, line 85: `op[:2]` always succeeds, so you don't actually need the length check. Not that it much matters, just python trivia.

clvm/operators.py
```diff
@@ -35,6 +49,97 @@
...
+        raise EvalError("reserved operator", args.to(op))
+
+    # all other unknown opcodes are no-ops
+    # the cost of the no-ops is determined by the opcode number, except the
+    # least significant byte. unknown opcodes of 1 byte have cost 1.
+    if len(op) <= 1:
+        return (1, SExp.null())
```
richardkiss, line 92: We might want some variance in 1-byte opcodes, since they are the most useful to repurpose (since they are small in the stream, especially if < 128).

clvm/operators.py
richardkiss: The return value of a `return` expression is None. :P

clvm/operators.py
aqk: Thanks, that's good Python info. Maybe a gotcha if coming from C++ land.

richardkiss (approved):
Looks good to me.

clvm/operators.py
```diff
@@ -35,6 +49,97 @@
...
+    cost_function = op[-2] & 0x3
+    # the multiplier cannot be 0. it starts at 1
+    cost_multiplier = (int.from_bytes(op[:-1], "big", signed=False) >> 2) + 1
+
+    # 0 = constant
+    # 1 = like op_add/op_sub
+    # 2 = like op_multiply
+    # 3 = like op_concat
+    if cost_function == 0:
+        cost = 1
```
aqk, line 103: > cost = 1

I believe we have a MIN_COST global floating around somewhere. May want to use that instead of `1`

clvm/operators.py
arvidn: you need to squint pretty hard :)

clvm/operators.py
arvidn: I wanted to keep this minimum cost to the actual absolute minimum, to leave the most doors open.

clvm/operators.py
arvidn: yeah, I seem to recall someone suggesting a full byte, but I may be mistaken. Maybe it would be better to move the 2 bits of `cost_function` into the last byte, and just have 6 bits unused for purposes of computing costs.

clvm/operators.py
arvidn: I've addressed this. I made the first condition check whether `len(op) == 0` instead. I don't think it makes sense to allow an opcode to be empty

arvidn:
All of the tests were auto-generated. I wanted to make sure all cases were covered so we can run the same tests against `clvm_rs` later to ensure it has the same costs.

arvidn:
This is ready for review, but before it lands I want to squash it and possibly undo the `clvm_tools` dependency (but we'd need to release `clvm_tools` first in that case)

arvidn:
one problem with this definition of cost is that it's unbounded. I think that's a problem. At least in the rust implementation we assume cost fits in a `uint32`. Any suggestions on how to solve this?

I principle, cost is also unbounded by the arguments we pass, so rust needs to be able to deal with this either way. Maybe it's not important, but we would need a rule to say if cost equal or exceeds 2^32, it's a fatal error

arvidn:
this is what I propose: https://github.com/Chia-Network/clvm/pull/60/commits/0236e265ce7ec8efb22422fd8f4297a920d78001

clvm/operators.py
```diff
@@ -35,6 +49,103 @@
...
+        for arg in args.as_iter():
+            if arg.pair:
+                raise EvalError("unknown op on list", arg)
+            cost += CONCAT_COST_PER_ARG
+            length += len(arg.atom)
+        cost += length // CONCAT_COST_PER_BYTE_DIVIDER
+
+    cost *= cost_multiplier
+    if cost >= 2**32:
+        raise EvalError("invalid operator", args.to(op))
```
arvidn, line 144: I think this restriction is very reasonable. A consequence of it is that any unknown operator that are longer than 5 (non-zero) bytes will also be a hard failure, since the `cost_multiplier` would necessarily cause the cost to exceed this limit.

PR #62, opened 2021-02-06
force the trace tests to run with the python back-end

arvidn:
The rust implementation does not support the pre-eval hooks. This change lets the rust implementation use these same test cases unmodified


arvidn:
I suppose this depends on a new release of `clvm_tools`

aqk (approved):

PR #64, opened 2021-02-08
OperatorDict test

aqk:


tests/operatordict_test.py
```diff
@@ -0,0 +1,29 @@
...
+        with self.assertRaises(AttributeError):
+            o = OperatorDict(d, apply=1)
+        with self.assertRaises(AttributeError):
+            o = OperatorDict(d, quote=1)
+        o = OperatorDict(d, apply=1, quote=2)
+        print(o)
+        # Why does the constructed Operator dict contain entries for "apply":1 and "quote":2 ?
+        # assert d == o
+        assert o.apply_atom == 1
+        assert o.quote_atom == 2
```
arvidn, line 24: how does this differ from `self.assertEqual()`? is the non-member assert to be preferred?

arvidn (approved):

tests/operatordict_test.py
richardkiss: Yeah, I'd recommend using `self.assertEqual` although I'm not exactly sure why.

aqk: self.assertEqual is a little better - display output on failure is different

PR #65, opened 2021-02-09
fix overlong encoding of zero in clvm serializer

arvidn:


tests/serialize_test.py
```diff
@@ -44,6 +44,14 @@ def check_serde(self, s):
         buf = sexp_buffer_from_stream(io.BytesIO(b))
         self.assertEqual(buf, b)
 
+    def test_zero(self):
```
aqk, line 47: Makes sense!

richardkiss (approved):

PR #67, opened 2021-02-15
Renumber keywords

richardkiss:
I've also renamed `div` to `/` for consistency. This will require a minor tweak to the rust error message (`div` => `/`) to pass tests there.


clvm/operators.py
```diff
@@ -23,11 +23,26 @@
 )
 
 KEYWORDS = (
-    ". q . a i c f r l x = sha256 + - * divmod "
-    "substr strlen point_add pubkey_for_exp concat . > >s "
-    "logand logior logxor lognot ash lsh "
-    "not any all "
-    "softfork div "
+    # core opcodes 0x01-x08
```
aqk, line 26: Ah, this is very nice.

clvm/operators.py
```diff
@@ -23,11 +23,26 @@
...
+    ". q a i c f r l x "
+
+    # opcodes on atoms as strings 0x09-0x0f
+    "= >s sha256 substr strlen concat . "
+
+    # opcodes on atoms as ints 0x10-0x17
+    "+ - * / divmod > ash lsh "
```
aqk, line 33: Might want to leave a blank spot in "opcodes on ints" namespace, just for aesthetics, if we add one later.

Hmm, come to think of it, will we specify that new opcodes are in a certain range?

arvidn (approved):

PR #68, opened 2021-02-16
add two more tests for substr

arvidn:


tests/brun/substr-d.txt
```diff
@@ -0,0 +1,2 @@
+brun -c '(substr (q "abcdefghijkl") 2 5)' (0 -1)'
```
aqk, line 1: How more there are 3 single quotes?

richardkiss (approved):

PR #71, opened 2021-02-23
make substr, lsh and ash parameters be int32

arvidn:
These are the last places where we currently don't charge for redundant leading zeros. ban wider than 4 bytes, even if they are leading zeros. This avoid having to charge cost for such extra bytes.


arvidn:
I actually meant to update the unknown op cost calculation as well, to use this same restriction. (the same way I did in the rust implementation [here](https://github.com/Chia-Network/clvm_rs/pull/52/files#diff-1106ab16843af039b2910d14818b53d52ab8542371d26e13f95b080b44617934L135))

clvm/operators.py
```diff
@@ -111,6 +111,10 @@ def default_unknown_op(op: bytes, args: CLVMObject) -> Tuple[int, CLVMObject]:
 
     cost_function = (op[-1] & 0b11000000) >> 6
     # the multiplier cannot be 0. it starts at 1
+
+    if len(op) > 5:
+        raise EvalError("invalid operator", args.to(op))
```
arvidn, line 116: the new requirements is that the cost multiplier may not be more than 4 bytes wide. It's still interpreted as an unsigned integer. This restriction was kind of implied before, by ensuring the total cost of the operation did not exceed `uint32_max`.

tests/unknown-op/unknown-00.txt
```diff
@@ -1,2 +1,2 @@
-brun --strict -c '(a (q 0x0000) (q ()))'
+brun --strict -c '(a (q . 0x000) (q . ()))'
```
arvidn, line 1: I'm not sure why these worked without having the quote updated. But I re-generated these to cover more interesting cases, with the maximum cost multiplier as well as invalid cost multipliers (5 bytes wide)

arvidn:
this was the program to re-generate the unknown op test cases:
```
cnt = 0
for args in ['(q . ())', '(q . (1 2 3))', '(q . (0x101010101010101010101010101010101010101010101010101010101010101010101010101010101010101101010101010101010101010101010101010101010101010101010101010101010101010101010101010101 0x101010101))', '(q . (1 2 (1 10 20)))']:
        for cost_fun in [0, 1, 2, 3]:
                for multiplier in [0, 1, 0xfffeffff, 0x7ffffffff]:
                        for ignored in [0, 63]:
                                for strict in [True, False]:
                                        open("unknown-%02d.txt" % cnt, "w+").write("brun " + ("--strict " if strict else "") + "-c '(a (q . 0x%x%02x) %s)'" % (multiplier, (cost_fun << 6) | ignored, args))
                                        cnt += 1

```

tests/unknown-op/unknown-00.txt
richardkiss: If you're changing these, can you change them to not use `a` (since its semantics are changing)? You can just go `(0x0000 (q . ()))` for example.

clvm/more_ops.py
```diff
@@ -68,6 +68,15 @@ def args_as_ints(op_name, args):
         yield (arg.as_int(), len(arg.as_atom()))
 
 
+def args_as_int32(op_name, args):
```
richardkiss, line 71: What do you think about moving this guy to `casts.py`?

richardkiss: Oh I guess that doesn't make sense, nm.

clvm/operators.py
richardkiss, line 116: We could just reset the cost in this case to something small, even a fixed value, rather than throw out the operator. I dunno, probably doesn't much matter. What do you think?

richardkiss (approved):
Looks great! The changes I suggested are optional, use your judgement as guide.

clvm/more_ops.py
```diff
@@ -261,7 +272,9 @@ def op_ash(args):
 
 
 def op_lsh(args):
-    (i0, l0), (i1, _) = args_as_int_list("lsh", args, 2)
+    (i0, l0), (i1, l1) = args_as_int_list("lsh", args, 2)
```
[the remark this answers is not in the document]
arvidn: the thing here is that the first parameter needs to be a full `BigInt`. It's just the second one that's an `int32`

clvm/operators.py
arvidn: I would think we have enough space for additional operators. If we feel we should have more space, I would think the lower bits (that are ignored for cost purposes) should be expanded.

arvidn:
I updated the unknown-operator tests to not use apply

clvm/more_ops.py
```diff
@@ -224,7 +233,7 @@ def op_substr(args):
     if a0.pair:
         raise EvalError("substr on list", a0)
 
-    (i1, _), (i2, _) = args_as_int_list("substr", args.rest(), 2)
+    i1, i2 = list(args_as_int32("substr", args.rest()))
```
aqk, line 236: Ah, good idea.

clvm/more_ops.py
aqk: `args_as_int32` could take an arg_count param

clvm/more_ops.py
```diff
@@ -248,7 +257,9 @@ def op_concat(args):
 
 
 def op_ash(args):
-    (i0, l0), (i1, _) = args_as_int_list("ash", args, 2)
+    (i0, l0), (i1, l1) = args_as_int_list("ash", args, 2)
+    if l1 > 4:
+        raise EvalError("ash requires int32 args (with no leading zeros)", args.rest().first())
```
aqk, line 262: I understand you can't use `args_as_int32` on the full arg list, but you might be able to centralize this check by:

>(i0, l0), (i1, l1) = args_as_int_list("ash", args, 2)
> i1 = args_as_int32("ash", args.rest(), 1)

Also redundant, unfortunately - No big deal
:)

clvm/operators.py
```diff
@@ -77,15 +77,15 @@ def args_len(op_name, args):
...
-# n | .... | 1          | 0          |
-# --+- - --+------------+------------+
-# n | .... |            |XX | XXXXXX |
-# --+- - --+------------+---+--------+
-# ^                      ^   ^
-# |                      |   + 6 bits ignored when computing cost
-# cost_multiplier        |
-#                        + 2 bits
-#                          cost_function
+# | 4 | 3 | 2 | 1 | 0          |
```
aqk, line 80: Oh, that's more readable, thanks!

clvm/more_ops.py
arvidn: note that this is actually a generator, so it would be a complicated place to check number of arguments. However, this is only used after the number of arguments are already checked

clvm/operators.py
aqk: Can you add the comment that it's still interpreted as an unsigned integer?
Maybe that is discussed during cost computation.

tests/brun/ash-c.txt
```diff
@@ -1,3 +1,2 @@
 brun -c '(ash (q . 0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000007) (q . 0x0000000000000000000000000000000000000000000000000000000000000001))'
-cost = 76
-14
+FAIL: ash requires int32 args (with no leading zeros) 0x0000000000000000000000000000000000000000000000000000000000000001
```
aqk, line 2: These new size restrictions for args 2&3 of substr and arg 2 of ash & lsh are meant to apply in all modes (safe & develop), right?

That seems correct to me.

tests/brun/lsh-c.txt
```diff
@@ -1,3 +1,2 @@
 brun -c '(lsh (q . 0x00000000000000000000000000000000000000000000000000000000000000000007) (q . 0x000000000000000000000000000000000000000000000000000000000000000000001))'
-cost = 76
-14
+FAIL: lsh requires int32 args (with no leading zeros) 0x0000000000000000000000000000000000000000000000000000000000000000000001
```
aqk, line 2: Hmm, now that these fail because of arg size, we need new tests on the boundary.

clvm/operators.py
aqk: Can you add the comment that it's still interpreted as an unsigned integer?
Maybe that is discussed during cost computation.

tests/brun/ash-c.txt
arvidn: that's right

tests/brun/lsh-c.txt
arvidn: good point, I'll add those (in a new PR)

clvm/more_ops.py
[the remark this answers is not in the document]
richardkiss: Oh yeah, ugh!

PR #72, opened 2021-02-24
Simplify `a`

richardkiss:


clvm/run_program.py
```diff
@@ -157,16 +157,10 @@ def apply_op(op_stack: OpStackType, value_stack: ValStackType) -> int:
...
-            new_operand_list = operand_list.rest().first()
-            if new_operator.pair:
-                new_pair = new_operator.cons(new_operand_list)
-                value_stack.append(new_pair)
-                op_stack.append(eval_op)
-            else:
-                value_stack.append(new_operator)
-                value_stack.append(new_operand_list)
-                op_stack.append(apply_op)
+            new_program = operand_list.first()
```
aqk, line 160: Yeah!

tests/brun/apply-00.txt
```diff
@@ -1,3 +1,3 @@
-brun -c '(a (q . +) (q . (3 4 5)))'
-cost = 84
-12
+brun -c '(a (q . 2) (q . (3 4 5)))'
```
aqk, line 1: I'm of two mind using the opcodes vs the keyword here.

It's less ambiguous for consensus critical interpretation of clvm byte code to use the opcodes.

But it makes it harder to read the test. Hmm, that gives me an idea for a small set of new tests.

tests/brun/apply-07.txt
```diff
@@ -1,3 +1,3 @@
-# a surprising case
-brun --strict -c '(a (q . q) (q . (2 3 4)))'
-FAIL: unimplemented operator 1
+brun --strict -c -n '(a (q . q) (q . (2 3 4)))'
+cost = 57
+(2 3 4)
```
aqk, line 3: Okay, that's nice.

tests/brun/max-cost-1.txt
```diff
@@ -1,3 +1,3 @@
-brun -c -m 75 '((c (q . 1) (q . (100 200))))'
-cost = 75
-(100 200)
+brun -c -m 73 '(c (q . 100) (q . 200))'
```
aqk, line 1: I think you should add a test to see `((c ...` fail

aqk (changes requested):
Mainly to add a test to see `((c ...` fail, if we don't have one yet.

tests/brun/apply-00.txt
richardkiss: The change here is to make explicit that `+` is no longer treated as an operator, but a path.

aqk (approved):

tests/brun/max-cost-1.txt
richardkiss: Done.

richardkiss: (Except it doesn't yet fail, that comes after a spin of clvm_tools.)

PR #73, opened 2021-02-25
The `((...))` construct is now a clvm error

richardkiss:


tests/brun/max-cost-2.txt
```diff
@@ -1,2 +1,2 @@
 brun -m 72 '(c (q . 100) (q . 200))'
-FAIL: cost exceeded 72
+FAIL: cost exceeded 26
```
richardkiss, line 2: This is because of a gross hack I had to make in `clvm_tools` to keep cost appearing the same (there's an offset to `max_cost` before `run_program` is invoked). I promise this will all be made less gross very soon.

aqk (approved):

PR #74, opened 2021-03-01
a few more test cases for lsh

arvidn:
demonstrating stripping leading zeros and adding a leading zero to ensure the result is unsigned

### DIFF
--- a/tests/brun/lsh-d.txt
+++ b/tests/brun/lsh-d.txt
@@ -1,0 +1,3 @@
+brun -c '(lsh (q . 0xffffff) (q . 0))'
+cost = 76
+0x00ffffff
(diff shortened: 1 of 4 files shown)


tests/brun/lsh-d.txt
```diff
@@ -0,0 +1,3 @@
+brun -c '(lsh (q . 0xffffff) (q . 0))'
+cost = 76
+0x00ffffff
```
aqk, line 3: Ah, yes, that's a good one. I'll look and see if we have the same for ash

aqk: I see this one

> edge-cases/ash-06.txt
> 1:brun '(ash (q . 0xffff) (q . 0xffff))'

aqk (approved):

PR #75, opened 2021-03-03
Add double-cons `((+) 1 2 3)` => `6` syntax.

richardkiss:
This is proposal for what `((X) ...)` should work. It shouldn't affect the compiler much because it never generates this construct. The optimizer could be upgraded to recognize that it's a constant and evaluate it at compile-time. It would allow you to more concisely write "add up this list of numbers"... if you have a list of numbers at node `5`, you could write

`(a (c (c (q . #+) 0) 1) 5)`

which is more compact than writing

`(a (c (q . #+) (....a recursive function that quotes each item in the list in node 5...)) 1)`.

(There's nothing magic about `+`... any operator besides `q` and `a` should work.)


richardkiss:
I'll draw up a clvm-rs version too.

tests/brun/cons-as-op-1.txt
```diff
@@ -1,2 +1,2 @@
 brun -c '((c (q . (+ (q . 50) 1)) (q . 500)))'
-FAIL: (()) eval disabled ((c (q 16 (q . 50) 1) (q . 500)))
+FAIL: in ((X)...) syntax X must be lone atom ((c (q 16 (q . 50) 1) (q . 500)))
```
aqk, line 2: > in ((X)...) syntax X must be

A comma might make this more readable:

in ((X)...) syntax, X must be ...

aqk (approved):

PR #76, opened 2021-03-04
Update to new version of blspy and use PrivateKey to multiply

mariano54:
PrivateKey should be used for representing ints, and constructor for infinity element


hoffmang9 (approved):
lgtm

wjblanke (approved):
aok

PR #77, opened 2021-03-09
Add substr_r (right substr) opcode, that returns a substring from the…

aqk:
… given index to the end of the atom (as a string)

### DIFF
--- a/clvm/more_ops.py
+++ b/clvm/more_ops.py
@@ -244,6 +244,24 @@ def op_substr(args):
     return cost, args.to(s)
 
 
+def op_substr_r(args):
+    if args.list_len() != 2:
+        raise EvalError("substr_r takes exactly 2 arguments", args)
+    a0 = args.first()
+    if a0.pair:
+        raise EvalError("substr_r on list", a0)
+
+    start_index = list(args_as_int32("substr_r", args.rest()))[0]
+
+    s0 = a0.as_atom()
+    end_index = len(s0)
+    if end_index < start_index or start_index < 0:
+        raise EvalError("invalid indices for substr_r", args)
+    s = s0[start_index:end_index]
+    cost = 1
+    return cost, args.to(s)
+
+
 def op_concat(args):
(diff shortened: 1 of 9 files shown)


aqk:
Hey everyone, Richard suggested this originally as a single-argument version of `substr` (which makes plenty of sense, especially since the remaining argument is the first argument of `substr`.

This function is known as `RIGHT$` in some BASICs. I thought I remembered C++ adding a good name for this, too, but I couldn't find it.

I'm open to changing this is pretty much any way, as long as we all agree. I think it is a useful operation.

Also, if you can think of a better name, that's cool, too.

clvm/more_ops.py
```diff
@@ -244,6 +244,24 @@ def op_substr(args):
...
+        raise EvalError("substr_r on list", a0)
+
+    start_index = list(args_as_int32("substr_r", args.rest()))[0]
+
+    s0 = a0.as_atom()
+    end_index = len(s0)
+    if end_index < start_index or start_index < 0:
+        raise EvalError("invalid indices for substr_r", args)
+    s = s0[start_index:end_index]
+    cost = 1
```
arvidn, line 261: I think it might make sense to use the same cost function as `substr`. I think we charge a little bit per byte, no?

arvidn:
the index parameter is an `int32`, meaning it's not allowed to be more than 4 bytes (even if those bytes are just leading zeroes). This is to avoid having complicated rules for how to charge cost per byte for "overlong" integer encodings.

We should have tests for indices that are longer than 4 bytes, both values that are too large but also leading zeroes.

clvm/more_ops.py
aqk: I was surprise to see `cost = 1` in substr, above, but I figured that was being worked on.

I knew you had that under control. I should have mentioned it, though.

richardkiss:
Replaced with https://github.com/Chia-Network/clvm/pull/80

PR #79, opened 2021-03-18
update costs to be more RPi friendly

arvidn:


clvm/costs.py
```diff
@@ -22,16 +22,16 @@
...
-SHA256_COST_PER_BYTE_DIVIDER = 64
+SHA256_COST_PER_ARG = 2
+SHA256_COST_PER_BYTE_DIVIDER = 16
 
 POINT_ADD_BASE_COST = 316
-POINT_ADD_COST_PER_ARG = 4200
+POINT_ADD_COST_PER_ARG = 23000
 
-PUBKEY_BASE_COST = 4194
+PUBKEY_BASE_COST = 23000
```
arvidn, line 34: this is the major change. `point_add` and `pubkey_for_exp` become significantly more expensive.

clvm/costs.py
```diff
@@ -5,11 +5,11 @@
 LISTP_COST = 1
 
 ARITH_BASE_COST = 1
-ARITH_COST_PER_LIMB_DIVIDER = 32
+ARITH_COST_PER_LIMB_DIVIDER = 16
```
arvidn, line 8: All these dividers mean how many bytes of atom you need to pay 1 in cost. So, lower divider means you pay more for memory. Some of these were set a bit high even for powerful computers, but especially for RPi where memory is expensive.

I think it still makes sense since most normal-sized integers will still be free from a cost-per-byte perspective.

clvm/costs.py
aqk: It seems that higher cost for `point_add` and  `pubkey_for_exp` is probably appropriate, even if it wasn't actual fact on the Raspberry Pi.

clvm/costs.py
aqk: Thanks - I recognize the limb nomenclature from GMP - do you think the use is complicated enough to warrant a comment?

tests/brun/pubkey-for-exp-3.txt
```diff
@@ -1,3 +1,3 @@
 brun -c '(pubkey_for_exp (q . 3))'
-cost = 4248
+cost = 23054
```
aqk, line 2: Hmm, this makes me think - what integer type to we use to hold the cost in the Rust code?

aqk (approved):

clvm/costs.py
richardkiss: We should probably discuss further. I know RP4 is supposed to be the "official" target, but this suggests maybe it shouldn't be. Maybe ARM isn't as well optimized as x86?

tests/brun/pubkey-for-exp-3.txt
arvidn: we use `u64`

clvm/costs.py
arvidn: something I've been wanting to do is to rename `LIMB` to `BYTE`, because we actually charge for leading zero bytes now, which technically don't turn into limbs

clvm/costs.py
arvidn: > Maybe ARM isn't as well optimized as x86?

possibly

aqk (approved):
This is probably a lot more representative of most nodes.

clvm/costs.py
richardkiss: I'm fine with the `LIMB` to `BYTE` change. Feel free to include it in this change if you want.

richardkiss (approved):

PR #80, opened 2021-04-01
First crack at two-parameter `substr`.

richardkiss:


tests/brun/substr_r-1.txt
```diff
@@ -0,0 +1,3 @@
+brun -c '(substr (q . "abcdefghijkl") (q . 9) (q . 10))'
```
arvidn, line 1: this looks like a duplicate test, taking two indices

arvidn (approved):

aqk (approved):

tests/brun/substr_r-1.txt
richardkiss: I removed it, and another one that is 3-arg.

richardkiss: Also renamed/renumbered all the `substr` tests in `brun`.

PR #85, opened 2021-04-20
Update CLVMObject.py

b0mTrady:
Added tiny amount of documentation for __slots__ dunder and typing.Optional for context


arvidn:
I don't think this documentation belongs here.

PR #91, opened 2021-05-29
add some type annotations

arvidn:
This is in preparation for making `SExp` have a clearer API and not derive from `CLVMObject`

### DIFF
--- a/clvm/core_ops.py
+++ b/clvm/core_ops.py
@@ -1,4 +1,5 @@
 from .EvalError import EvalError
+from .SExp import SExp
 
 from .costs import (
     IF_COST,
@@ -11,7 +12,7 @@ from .costs import (
 )
 
 
-def op_if(args):
+def op_if(args: SExp):
     if args.list_len() != 3:
         raise EvalError("i takes exactly 3 arguments", args)
     r = args.rest()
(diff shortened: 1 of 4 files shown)


clvm/core_ops.py
```diff
@@ -20,41 +21,41 @@ def op_if(args):
...
 
-def op_eq(args):
+def op_eq(args: SExp):
     if args.list_len() != 2:
         raise EvalError("= takes exactly 2 arguments", args)
     a0 = args.first()
     a1 = args.rest().first()
-    if a0.pair or a1.pair:
-        raise EvalError("= on list", a0 if a0.pair else a1)
+    if a0.as_pair() or a1.as_pair():
```
richardkiss, line 57: I think we should require that `SExp` also implement the `CLVMObject` protocol, ie. respond to `.pair` and `.atom` which makes this change unnecessary.

arvidn: I think it's problematic to have `SExp` also implement the `CLVMObject` protocol, as we can end up nesting them which makes things a bit more confused.

I'm fine with `SExp` having a different API, but I don't see a good reason to have both `as_pair()` *and* just `pair`

clvm/more_ops.py
```diff
@@ -49,16 +50,16 @@
 )
 
 
-def malloc_cost(cost, atom):
-    return cost + len(atom.atom) * MALLOC_COST_PER_BYTE, atom
+def malloc_cost(cost, atom: SExp):
+    return cost + len(atom.as_atom()) * MALLOC_COST_PER_BYTE, atom
```
richardkiss, line 54: Ditto this, can be reverted. (I used to use `.as_atom` and `.as_pair` exclusively, but decided to switch to using `.atom` and `.pair` when possible so that we can use `CLVMObject` in more places.)

richardkiss (changes requested):
I support adding the additional typing information. But I still would like to use the preferred and faster `.atom`/`.pair` API when possible.

arvidn:
is your opinion that `SExp`'s API should be `.atom` and `.pair` properties instead of `.as_atom()` and `.as_pair()`, or that it should have both?

arvidn:
I think the fact that `CLVMObject` and `SExp` share some part of their interfaces is problematic, and makes it more difficult to separate their responsibilities and uses.

richardkiss:
Both. We can have `SExp` be a subclass of `CLVMObject`, but `.as_pair()` is subtly different from `.pair` in that it rewraps the pair of `CLVMObject` children as `SExp` objects.

`SExp` is basically so you can get convenience features like `.as_iter` and `.as_int`. It really shows how clever rust `impl` blocks are, since it'd be much nicer to just be able to add these methods right into `CLVMObject`.

richardkiss (approved):

PR #94, opened 2021-06-01
make SExp wrap any tree implementing the CLVMObject protocol.

arvidn:
Make CLVMObject lazily convert python types to CLVM types as elements are inspected.


clvm/CLVMObject.py
```diff
@@ -1,47 +1,107 @@
...
-                and len(v) == 2
-                and isinstance(v[0], CLVMObject)
-                and isinstance(v[1], CLVMObject)
-            ) or isinstance(v, bytes)
-            # uncomment next line for debugging help
-            # if not type_ok: breakpoint()
-            assert type_ok
+
+        # TODO: ideally we wouldn't mix SExp into the tree like this
+        if type(v).__name__ in ["SExp", "Program"]:
```
arvidn, line 45: is there a way to use `isinstance()` here, without importing `SExp`? As far as I understand it's not possible to import it here because you end up with a circular dependency

richardkiss: Yeah, hard-coding `SExp` and `Program` here seems wrong. It might be possible to use `register` on an abstract base class as described here https://docs.python.org/3/library/abc.html but I haven't tested this.

arvidn: I think the source of this issue is that we create `CLVMObjects` from `SExp` sometimes, so we need to unwrap them here. if we would always use the higher level `SExp` to build trees, we would be fine since it can do the unwrapping. So I consider this temporary measure regardless

clvm/ConvertingCLVMObject.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.atom_
+
+    @property
+    def pair(self):
+        if self.pair_ is not None:
+            valid_first = isinstance(self.pair_[0], ConvertingCLVMObject) or looks_like_clvm_object(self.pair_[0])
+            valid_second = isinstance(self.pair_[1], ConvertingCLVMObject) or looks_like_clvm_object(self.pair_[1])
+            if not valid_first or not valid_second:
+                # deferred conversion
+                self.pair_ = (self.pair_[0] if valid_first else ConvertingCLVMObject(self.pair_[0]),
```
arvidn, line 85: maybe there's a way to assign `self.pair` in here, and `self.atom` in the other property, that would make any future attempts to access these fields just go straight to the members, rather than these functions.

prozacchiwawa (approved):
Neat.

clvm/ConvertingCLVMObject.py
richardkiss:
```python
class foo:
    @property
    def val(self):
        self.val = 100
        return 200

f1 = foo()
print(f1.val)
print(f1.val)
print(f1.val)
```

This prints
```
200
100
100
100
```

so there's your hack.

clvm/ConvertingCLVMObject.py
```diff
@@ -0,0 +1,103 @@
...
+    # in CLVMObject to defer conversion when accessed
+    pair_: Optional[Tuple[Any, Any]]
+    __slots__ = ["atom", "pair_"]
+
+    def __new__(class_, v: CastableType):
+        if isinstance(v, ConvertingCLVMObject):
+            return v
+
+        # TODO: ideally we wouldn't mix SExp into the tree like this
+        if type(v).__name__ in ["SExp", "Program"]:
```
richardkiss, line 47: Maybe just do `hasattr(v, "obj_")` ?

clvm/ConvertingCLVMObject.py
arvidn: I can't get this to work. I've tried `self.pair = `, `setattr(self, "pair", ...)`, `super().__setattr__("pair", ...)`. They all fail with:
```
AttributeError: can't set attribute
```

Presumably because `pair` is still treated as a property that doesn't have a setter, just a getter.

The one thing that didn't fail was `self.__dict__["pair"] =`, but that did not override the `@property`, so it didn't do any good.

I imagine that perhaps this is related to `__slots__ = ["atom", "pair_"]`. That was one of the first things I removed, to enable "objects are arbitrary dicts" behavior. I also suspected it had something to do with the existence if `__new__()`. I tried to replace that with an `__init__()`, but that also did not make a difference.

richardkiss: It turns out I was (unwittingly) running it under `python2`.

arvidn:
I updated https://github.com/Chia-Network/chia-blockchain/pull/6431 to use the most recent version of this patch, to ensure all chia-blockchain tests still pass.

clvm/CLVMObject.py
```diff
@@ -1,47 +1,32 @@
...
-            # uncomment next line for debugging help
-            # if not type_ok: breakpoint()
-            assert type_ok
         self = super(CLVMObject, class_).__new__(class_)
         if isinstance(v, tuple):
+            if len(v) != 2:
+                raise ValueError("tuples must be of size 2, cannot create CLVMObject from: %s" % str(v))
             self.pair = v
             self.atom = None
+        elif v is None:
```
richardkiss, line 26: What's the motivation for adding the `CLVMObject(None)` case?

arvidn:
also updated: https://github.com/Chia-Network/chia-blockchain/pull/6431

richardkiss (approved):

PR #98, opened 2021-07-01
My take on dialects, which I think is now uniform in interface with the run_program from stage_0

prozacchiwawa:
May be controversial.

I made cuts needed to move some constants into a place where importing them would be possible from anywhere, improving dependency congestion.

Main visible differences:

- There is now a dialect_factories collection the end user can use to get sane dialects.
- Configuring a dialect to be a standard chia clvm runner is now a function that most python based dialects can use.
- A configure method on dialects allows implementation specific communication

As used in clvm_tools, it's now possible to

            ````
            backend = args.backend if args.backend is not None else "python"
            dialect = dialect_factories[backend](
                KEYWORD_TO_ATOM["q"],
                KEYWORD_TO_ATOM["a"],
                args.strict,
                to_sexp_f
            )

            cost, result = dialect.run_program(
                run_script, input_sexp, max_cost=max_cost, pre_eval_f=pre_eval_f)
            ````

The dialect object is factored out of global paths, so it at least owns all its own information.

### DIFF
--- a/clvm/chia_dialect.py
+++ b/clvm/chia_dialect.py
@@ -1,37 +1,86 @@
+from .SExp import SExp
 from .casts import int_to_bytes
-from .dialect import ConversionFn, Dialect, new_dialect, opcode_table_for_backend
+from .types import CLVMObjectType, ConversionFn, MultiOpFn, OperatorDict
+from .chainable_multi_op_fn import ChainableMultiOpFn
+from .handle_unknown_op import (
+    handle_unknown_op_softfork_ready,
+    handle_unknown_op_strict,
+)
+from .dialect import ConversionFn, Dialect, new_dialect, opcode_table_for_backend, python_new_dialect, native_new_dialect
+from .chia_dialect_constants import KEYWORDS, KEYWORD_FROM_ATOM, KEYWORD_TO_ATOM  # noqa
+from .operators import OPERATOR_LOOKUP
 
-KEYWORDS = (
-    # core opcodes 0x01-x08
-    ". q a i c f r l x "
-
-    # opcodes on atoms as strings 0x09-0x0f
-    "= >s sha256 substr strlen concat . "
-
-    # opcodes on atoms as ints 0x10-0x17
-    "+ - * / divmod > ash lsh "
-
-    # opcodes on atoms as vectors of bools 0x18-0x1c
-    "logand logior logxor lognot . "
-
-    # opcodes for bls 1381 0x1d-0x1f
-    "point_add pubkey_for_exp . "
-
-    # bool opcodes 0x20-0x23
-    "not any all . "
-
-    # misc 0x24
-    "softfork "
-).split()
-
-KEYWORD_FROM_ATOM = {int_to_bytes(k): v for k, v in enumerate(KEYWORDS)}
-KEYWORD_TO_ATOM = {v: k for k, v in KEYWORD_FROM_ATOM.items()}
-
-
-def chia_dialect(strict: bool, to_python: ConversionFn, backend=None) -> Dialect:
+def configure_chia_dialect(dialect: Dialect, backend=None) -> Dialect:
     quote_kw = KEYWORD_TO_ATOM["q"]
     apply_kw = KEYWORD_TO_ATOM["a"]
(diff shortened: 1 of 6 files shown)


clvm/chia_dialect.py
```diff
@@ -0,0 +1,86 @@
...
+    ):
+        super().__init__(quote_kw, apply_kw, multi_op_fn, to_python)
+        self.tracer = lambda x,y: None
+
+    def do_sha256_with_trace(self,prev):
+        def _run(value,max_cost=None):
+            try:
+                cost, result = prev(value)
+            except TypeError:
+                cost, result = prev(value,max_cost)
```
richardkiss, line 41: Can you run `black` on all the `.py` files to standardize formatting?

```
pip install black
black clvm tests
```

clvm/chia_dialect.py
```diff
@@ -0,0 +1,86 @@
...
+# Dialect that can allow acausal tracing of sha256 hashes.
+def debug_new_dialect(
+    quote_kw: bytes, apply_kw: bytes, strict: bool, to_python: ConversionFn,
+    backend="python"
+) -> Dialect:
+    d = chia_python_new_dialect(quote_kw, apply_kw, strict, to_python, backend)
+
+    # Override operators we want to track.
+    std_op_table = opcode_table_for_backend(KEYWORD_TO_ATOM, backend="python")
+    table = { b'\x0b': d.do_sha256_with_trace(std_op_table[b'\x0b']) }
```
richardkiss, line 77: Maybe use `KEYWORD_TO_ATOM["sha256"]` instead of hard-coding `0x0b`.

clvm/chia_dialect_constants.py
```diff
@@ -0,0 +1,43 @@
...
+    "not any all . "
+
+    # misc 0x24
+    "softfork "
+).split()
+
+KEYWORD_FROM_ATOM = {int_to_bytes(k): v for k, v in enumerate(KEYWORDS)}
+KEYWORD_TO_ATOM = {v: k for k, v in KEYWORD_FROM_ATOM.items()}
+
+KEYWORD_TO_LONG_KEYWORD = {
```
richardkiss, line 29: Can you build this table algorithmically from `OP_REWRITE`?

richardkiss: Or vice-versa?

richardkiss: I'm not sure this constant is actually Chia dialect-specific.

setup.py
```diff
@@ -7,6 +7,7 @@
 
 dependencies = [
     "blspy>=0.9",
+    "clvm_rs>=0.1.8"
```
richardkiss, line 10: I think it's best to make `clvm_rs` an optional dependency. I expect that `clvm_rs` will eventually be completely standalone.

clvm/types.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any, Callable, Dict, Tuple, Union
+
+
+CLVMAtom = Any
+CLVMPair = Any
+
+CLVMObjectType = Union["CLVMAtom", "CLVMPair"]
```
richardkiss, line 7: I know I originally wrote this file, but I'm wondering if maybe we should remove `CLVMObjectType` since it's not actually used anywhere (and maybe other unused things in this file). It doesn't seem that mypy's typing system is advanced enough to do what we need to do.

clvm/chia_dialect.py
prozacchiwawa: makes sense

clvm/chia_dialect_constants.py
prozacchiwawa: I included it there mainly to avoid another layer when resolving dependencies, but it does make sense to put this on a layer in between.

tests/brun/trace-2.txt
```diff
@@ -1,4 +1,4 @@
-brun --backend=python -c -v '(x)'
+brun -c -v '(x)'
```
prozacchiwawa, line 1: Whoops these should be replaced with a dialect choice.  Will fix that.

PR #99, opened 2021-07-29
Refactor for `Dialect`.

richardkiss:


tests/brun/trace-2.txt
```diff
@@ -1,4 +1,4 @@
-brun --backend=python -c -v '(x)'
+brun -c -v '(x)'
```
arvidn, line 1: this would cause the `clvm_rs` tests to fail, wouldn't it?

clvm/dialect.py
```diff
@@ -0,0 +1,168 @@
...
+        to_python: ConversionFn,
+    ):
+        self.quote_kw = quote_kw
+        self.apply_kw = apply_kw
+        self.opcode_lookup = dict()
+        self.multi_op_fn = ChainableMultiOpFn(self.opcode_lookup, multi_op_fn)
+        self.to_python = to_python
+
+    def update(self, d: OperatorDict) -> None:
+        self.opcode_lookup.update(d)
```
arvidn, line 100: given that you have this `update()` function, why do you need to be able to chain the `multi_op_fn` with `ChainableMultiOpFn`? it looks more complex than it needs to be

clvm/dialect.py
```diff
@@ -0,0 +1,168 @@
...
+            pre_eval_f,
+        )
+        return cost, self.to_python(r)
+
+
+def native_new_dialect(
+    quote_kw: bytes, apply_kw: bytes, strict: bool, to_python: ConversionFn
+) -> Dialect:
+    unknown_op_callback = (
+        clvm_rs.NATIVE_OP_UNKNOWN_STRICT
```
arvidn, line 130: are these flags or magic-values exported by the rust bindings, to indicate it's a native rust function?

tests/operatordict_test.py
```diff
@@ -1,29 +0,0 @@
...
-        print(o)
-        # Why does the constructed Operator dict contain entries for "apply":1 and "quote":2 ?
-        # assert d == o
-        self.assertEqual(o.apply_atom, 1)
-        self.assertEqual(o.quote_atom, 2)
-
-        # Test construction from an already existing OperatorDict
-        o2 = OperatorDict(o)
-        self.assertEqual(o2.apply_atom, 1)
-        self.assertEqual(o2.quote_atom, 2)
```
arvidn, line 29: `OperatorDict` is still a type that exists. It's not clear why it shouldn't have a test

PR #102, opened 2021-08-10
Added py.typed files to packages so that mypy can analyze them

Quexington:


Quexington:
This PR is technically good to go in terms of this repository, the problem is that since the types will be detectable now, the `chia-blockchain` repository is going to face a bunch of linting errors.  Best hold off until there is a corresponding PR to fix those as well.

altendky:
Unfixed errors are being ignored in https://github.com/Chia-Network/chia-blockchain/pull/9371.

mariano54:
We should merge this imo @Quexington, dealing with the issues will be a good thing

altendky:
I had done some more hinting work, but I don't really recall the form and now it has become disconnected seemingly.  https://github.com/Chia-Network/clvm/pull/107  Maybe next week I could look back around this stuff?  Maybe I'm being overly optimistic about that?  `:|`

Quexington:
I agree @mariano54 it just would have required a rather large PR into chia-blockchain (and certainly a larger one now).  One of the biggest fixes will be that `SerializedProgram` and `Program` will actually be distinct and there's a TON of places where we pass a `Program` where it's expecting the latter IIRC.

PR #112, opened 2022-03-04
update the expected cost from brun calls in test cases

arvidn:
To accomodate for the change in how brun reports cost.

Once https://github.com/Chia-Network/clvm_tools/pull/82 lands, the CI for this should be re-triggered.


richardkiss (approved):
Looks like everything goes down by 9. Seems good. Probably we should merge clvm_tools first, do a release, then pin that newly released version here in `setup.py`. That's the only change I'd request.

PR #113, opened 2022-03-24
make op_div behave like mempool and post soft-fork

arvidn:


setup.py
```diff
@@ -10,7 +10,7 @@
 ]
 
 dev_dependencies = [
-    "clvm_tools>=0.4.2",
+    "clvm_tools>=0.4.4",
```
arvidn, line 13: I believe the tests rely on the most recent version of `clvm_tools` now, since we changed the cost reporting

richardkiss (commented):
Seems fine. I'd really like to have dialect support in python clvm, but that's kind of out of scope of this PR. Would you consider maybe just creating a new version of `op_div` and renaming the existing one `op_div_original` or something? That way if and when we get around to adding dialect support to `clvm` we can resurrect the old behaviour as a particular (admittedly fairly useless except maybe for historical forensics) dialect.

arvidn:
I don't see a lot of value in preserving the old behaviour. We don't use it. If someone will wants that behaviour in the future, it's easy to add  back.

PR #115, opened 2022-04-22
Add macOS to CI

altendky:


altendky:
In chia-blockchain we exclude macOS with Python 3.7.  I don't know if it adds any value at this point to add that exclusion.  It's easy enough in the matrix if we want it.

richardkiss (approved):

PR #134, opened 2023-08-16
remove special case for `G1Element` in `SExp.to()`

arvidn:
Instead support any type that can convert to `bytes`.

### DIFF
--- a/clvm/SExp.py
+++ b/clvm/SExp.py
@@ -1,7 +1,6 @@
 import io
 import typing
 
-from blspy import G1Element
 
 from .as_python import as_python
 from .CLVMObject import CLVMObject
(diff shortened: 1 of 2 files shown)


richardkiss (commented):
More general and better. My only concern is that the type signature for `convert_atom_to_bytes` no longer reflects that `G1Element` is safe, so may cause `mypy` problems. This isn't easy to fix unless we ignore py3.7, which is EOL June, 2023, so maybe it's okay. ChatGPT says we can do something like this:
```
from typing import Protocol, runtime_checkable

@runtime_checkable
class CanBeCastToBytes(Protocol):
    def __bytes__(self) -> bytes: ...
```

but it only works in 3.8+.

richardkiss (approved):

arvidn:
yeah, as far as I can tell, in all situations we use this in `chia-blockchain` right now, the types of the list is erased already anyway. So in practice it's not a regression (yet). I suppose we would need to export a protocol that we could use in `chia-blockchain` as well, as the member of our `List[x]`

PR #144, opened 2023-12-22
Replace blspy with chia_rs

AmineKhaldi:


setup.py
```diff
@@ -6,7 +6,7 @@
     long_description = fh.read()
 
 dependencies = [
-    "blspy>=0.9",
+    "chia_rs==0.3.3",
```
altendky, line 9: Can we be a bit looser?  Otherwise this has to update in lock step with blockchain.

altendky (changes requested):

setup.py
AmineKhaldi: Done.

emlowe:
I think we could do with >=0.2.13 - that is the chia_rs version used by chia 2.1.2 where we switched to using it for BLS

AmineKhaldi:
> I think we could do with >=0.2.13 - that is the chia_rs version used by chia 2.1.2 where we switched to using it for BLS

Done.

richardkiss (approved):

PR #146, opened 2024-01-22
Add python 3.12

emlowe:
Add 3.12 to the test matrix.
Added `setuptools` as a dev dependency as a workaround for `clvm_tools` needing it until that repo can get cleaned up to not use `pkg_resources`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: [ '3.8', '3.9', '3.10', '3.11']
+        python-version: [ '3.8', '3.9', '3.10', '3.11', '3.12']
     steps:
       - uses: actions/checkout@v3
         with:
--- a/clvm/version.py
+++ b/clvm/version.py
@@ -1,6 +1,5 @@
 import importlib_metadata
 
-__version__: str
 try:
     __version__ = importlib_metadata.version(__name__)
 except importlib_metadata.PackageNotFoundError:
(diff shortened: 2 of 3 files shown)


emlowe:
close and reopen

richardkiss:
At first I didn't understand the `pkg_resources` stuff. There's another way around this that doesn't require changing `setup.py` at all:

```
diff --git a/clvm/version.py b/clvm/version.py
index 9e64dab..620de87 100644
--- a/clvm/version.py
+++ b/clvm/version.py
@@ -1,7 +1,7 @@
-from pkg_resources import get_distribution, DistributionNotFound
+from importlib.metadata import distribution, PackageNotFoundError
 
 try:
-    __version__ = get_distribution(__name__).version
-except DistributionNotFound:
+    __version__ = distribution(__name__.split(".")[0]).version
+except PackageNotFoundError:
     # package is not installed
     __version__ = "unknown"
```

Can you redo with this change? Or tell me to do it, and I'll hack the patch.

emlowe:
Go ahead Richard, but I think I needed `setuptools` because the current version of `clvm_tools` needed it  - but yea, go ahead hack away!

altendky:
https://github.com/Chia-Network/chia-blockchain/blob/fe89b5b782e3879c227820e86c1b16f1834726a0/chia/__init__.py

```python
from __future__ import annotations

import importlib.metadata

__version__: str
try:
    __version__ = importlib.metadata.version("chia-blockchain")
except importlib.metadata.PackageNotFoundError:
    # package is not installed
    __version__ = "unknown"
```

fwiw.  noting that the import package name doesn't have to agree with the distribution package name.  we like to vary them.

altendky:
Though, with it being this easy to get a package's version and it not being human readable looking at the source...  kinda makes you wonder what good `thepackage.__version__` is here, even if totally common to provide.  But, that debate can certainly be left for another time, if any of us care enough.

richardkiss (approved):

PR #168, opened 2025-03-21
Rebase and merge some long-ago written serialization-with-backrefs code.

richardkiss:


clvm/object_cache.py
```diff
@@ -0,0 +1,106 @@
...
+        If a pair is passed and one of the children does not have its T value cached
+        in `ObjectCache` yet, return `None` and f will be called with each child in turn.
+        Don't recurse in f; that's part of the point of this function.
+        """
+        self.f = f
+        self.lookup: Dict[int, Tuple[T, CLVMStorage]] = dict()
+
+    def get(self, obj: CLVMStorage) -> T:
+        obj_id = id(obj)
+        if obj_id not in self.lookup:
```
arvidn, line 39: technically, I don't think this condition isn't necessary.  it's repeated on line 44.

arvidn (approved):
I consider this more of a reference-implementation than a practical one.
It could probably do with some round-trip tests, including trees where nodes are reused.

clvm/object_cache.py
richardkiss: True enough. The counter is `.get` will be called frequently and this short-cut will prevent the list from being created in the more common "already cached" case.

richardkiss:
> I consider this more of a reference-implementation than a practical one. It could probably do with some round-trip tests, including trees where nodes are reused.

Agreed. It's a good baseline for some improvements I'm trying, in particular one that I believe turns an $n^2$ factor into an ${n} log {n}$ one.

PR #170, opened 2025-03-31
relax as_iter() to allow any atom to terminate the list

arvidn:


clvm/SExp.py
```diff
@@ -217,7 +217,7 @@ def null(class_) -> SExp:
 
     def as_iter(self: _T_SExp) -> typing.Iterator[_T_SExp]:
         v = self
-        while not v.nullp():
+        while v.listp():
```
richardkiss, line 220:
```suggestion
        while v.pair is not None:
```

richardkiss: At some point, I created `CLVMObject` which is supposed to be the "real" way to deal with the `Union` that is a clvm node: exactly one of `.atom` and `.pair` is `None`. I think it's better than using `.listp` (which should probably just go away).

arvidn: that was my first version of this :)

richardkiss (approved):

PR #173, opened 2025-06-17
Add an optional `max_size` parameter to serialize to prevent certain attacks where size can blow up

matt-o-how:
Adds an optional field to quit out of serializing if a maximum byte count is reached

This change also takes the time to update a couple of out of date things.

`py.test` is deprecated in favour of `pytest` and `mypy` is now catching a problem in more_ops.py

### DIFF
--- a/tests/serialize_test.py
+++ b/tests/serialize_test.py
@@ -8,6 +8,7 @@ import pytest
 from clvm import to_sexp_f
 from clvm.SExp import CastableType, SExp
 from clvm.serialize import (
+    MAX_SAFE_BYTES,
     _atom_from_stream,
     sexp_from_stream,
     sexp_buffer_from_stream,
@@ -146,14 +147,14 @@ class SerializeTest(unittest.TestCase):
             count = size // len(TEXT)
             text = TEXT * count
             assert len(text) < size
-            if len(text) >= 134217688:
+            if len(text) >= MAX_SAFE_BYTES:
                 with pytest.raises(ValueError, match="SExp exceeds maximum size"):
                     self.check_serde(text)
             else:
                 self.check_serde(text)
             text = TEXT * (count + 1)
             assert len(text) > size
-            if len(text) >= 134217688:
+            if len(text) >= MAX_SAFE_BYTES:
                 with pytest.raises(ValueError, match="SExp exceeds maximum size"):
                     self.check_serde(text)
             else:


clvm/serialize.py
```diff
@@ -26,6 +27,8 @@
 BACK_REFERENCE = 0xFE
 CONS_BOX_MARKER = 0xFF
 
+MAX_SAFE_BYTES = 2_000_000
```
arvidn, line 30: I don't see this constant being used anywhere. I think it would be reasonable to make it the default value of `max_size`

clvm/serialize.py
```diff
@@ -153,9 +156,14 @@ def atom_to_byte_iterator(as_atom: bytes) -> typing.Iterator[bytes]:
 
 
 def sexp_to_stream(
-    sexp: CLVMStorage, f: typing.BinaryIO, *, allow_backrefs: bool = False
+    sexp: CLVMStorage, f: typing.BinaryIO, *, allow_backrefs: bool = False, max_size: Optional[int] = None
 ) -> None:
+    remaining = max_size
```
arvidn, line 161: if you make `max_size` non-optional, you can just count down that variable instead of making a copy

tests/serialize_test.py
```diff
@@ -85,6 +92,11 @@ def check_serde(self, s: CastableType) -> bytes:
             self.assertEqual(v2, s)
             b3 = v2.as_bin()
             self.assertEqual(b, b3)
+        with pytest.raises(ValueError):
```
arvidn, line 95:
```suggestion
        with pytest.raises(ValueError, matches="SExp exceeds maximum size"):
```

tests/serialize_test.py
```diff
@@ -210,14 +230,18 @@ def make_bomb(depth: int) -> SExp:
         self.assertEqual(len(b10_2), 75)
 
         bomb_20 = make_bomb(20)
-        b20_1 = bomb_20.as_bin(allow_backrefs=False)
+        with pytest.raises(ValueError, match="SExp exceeds maximum size"):
+            # b20_1 = bomb_20.as_bin(allow_backrefs=False)
+            bomb_20.as_bin(allow_backrefs=False)
         b20_2 = bomb_20.as_bin(allow_backrefs=True)
-        self.assertEqual(len(b20_1), 48234495)
+        # self.assertEqual(len(b20_1), 48234495)
```
arvidn, line 237: this should probably be removed, unless you commented it out with the intention to do something else with it

tests/serialize_test.py
```diff
@@ -210,14 +230,18 @@ def make_bomb(depth: int) -> SExp:
         self.assertEqual(len(b10_2), 75)
 
         bomb_20 = make_bomb(20)
-        b20_1 = bomb_20.as_bin(allow_backrefs=False)
+        with pytest.raises(ValueError, match="SExp exceeds maximum size"):
+            # b20_1 = bomb_20.as_bin(allow_backrefs=False)
```
arvidn, line 234: this looks like it was left over by accident

tests/serialize_test.py
```diff
@@ -210,14 +230,18 @@ def make_bomb(depth: int) -> SExp:
...
         b20_2 = bomb_20.as_bin(allow_backrefs=True)
-        self.assertEqual(len(b20_1), 48234495)
+        # self.assertEqual(len(b20_1), 48234495)
         self.assertEqual(len(b20_2), 105)
 
         bomb_30 = make_bomb(30)
         # do not uncomment the next line unless you want to run out of memory
-        # b30_1 = bomb_30.as_bin(allow_backrefs=False)
+        with pytest.raises(ValueError, match="SExp exceeds maximum size"):
+            # b30_1 = bomb_30.as_bin(allow_backrefs=False)
```
arvidn, line 243: I think you can remove this commented-out line now, since you have a copy of it below, that throws

tests/serialize_test.py
```diff
@@ -134,10 +146,18 @@ def test_very_long_blobs(self) -> None:
             count = size // len(TEXT)
             text = TEXT * count
             assert len(text) < size
-            self.check_serde(text)
+            if len(text) >= 134217688:
```
richardkiss, line 149: What a very magic magic value!

clvm/serialize.py
richardkiss: Consider renaming it `DEFAULT_MAX_SAFE_BYTES`.

richardkiss:
An issue I can think of is that it makes life difficult for `clvm_tools` users who could work with ultra-large blobs before this. Now they're going to get serialization errors. I think we would have to add an option to `clvm_tools` to override this default max block size (which of course is a bit tricky because we have to coordinate to another repo).

richardkiss:
Or maybe just a new API with a new name which requires the `max_size` api, and we deprecate the old api. This is kind of like `Program.run` which originally had no limit, which proved to be stupid, so we had to move to a `Program.run_with_max_cost` or something like that, and then the `max_cost` eventually made its way into `.run`. Or at least in some places.

Conceptually it's fine. But I just worry about me getting dinged when trying to use `clvm_tools` at some point in the future. Maybe we do what we did with `Program` to give `clvm_tools` a chance to be revved too. Does anyone besides me even use it?

tests/serialize_test.py
matt-o-how: Replaced now with a non-magic value

arvidn (approved):

arvidn:
I think the current implementation is preferable implementing the limit in a stream adapter. We don't know of any use cases where it's OK to be unsafe (i.e. unlimited). even `clvm_tools` ought to have some limit. I don't think it's reasonable to provide an API that's unsafe, and even if we do, it should have a longer name indicating that it's unsafe.

PR #174, opened 2025-06-25
Add a max_size passthrough for SExp::as_bin

prozacchiwawa:
and add a test that verifies that it increases the amount of data that as_bin will tolerate.


Rigidity (approved):

richardkiss (approved):
Seem reasonable

arvidn (approved):

aqk (approved):

